PR #1, opened 2015-01-26
Fix gitignore, as needed after moving svir to the main folder in the new repo

ptormene:
Please, MB, check if pycharm-related files are correctly ignored now


mbernasocchi:
LGTM

PR #3, opened 2015-01-26
Remove button to copy loss data into zonal layer

ptormene:
It was redundant, because the same (and more) can be done by the button used to aggregate losses by zone


mbernasocchi:
lgtm

PR #4, opened 2015-01-26
Save downloaded data into shapefile

ptormene:
Necessary to be able to save a project correctly, otherwise the in-memory layer is not properly saved


mbernasocchi:
LGTM

PR #6, opened 2015-01-26
add close and show in browser button

mbernasocchi:
add a button that closes the uploader and shows the layer in the default system webbrowser


ptormene:
LGTM

PR #7, opened 2015-01-27
Add license to layer restrictions

ptormene:
Partially addressing https://bugs.launchpad.net/oq-svir/+bug/1409725
Note: unfortunately, the current version of Geonode does not put the license into a "license" metadata field, but into the "other restrictions" field. We can merge the current solution anyway, and wait for the next version of Geonode for further changes.


mbernasocchi:
lgtm

PR #9, opened 2015-01-27
Use POST request to export socioeconomic data from the platform

ptormene:
See https://bugs.launchpad.net/oq-svir/+bug/1414711 and the Platform-side companion PR https://github.com/gem/oq-platform/pull/363


mbernasocchi:
LGTM

PR #16, opened 2015-02-04
Set keywords as separate gmd:keyword elements in the metadata xml

ptormene:
(instead of using comma-separated keywords in a single element)


ptormene:
This is a very small change, and I've tested it, so I am merging it.

PR #17, opened 2015-02-09
Aggregate multivariate loss

ptormene:
When aggregating loss by zone, the user was able to select only one loss variable from the layer containing loss points. This PR adds the possibility to select multiple attributes and to perform the "aggregation by zone" just once, producing for each zone the count of points and the sum and average values for each of the loss attributes.

When aggregating losses using geometries, the user is enabled to select the attribute containing the zone id from the zonal layer (this is useful, because the aggregation produces a new version of the loss layer, with all points labelled with an additional attribute containing the zone id)

This PR also solves a corner case in which the new attribute that SAGA produces (labelling the point with the zone id) is already taken in the loss layer, and therefore the new attribute name is automatically modified. The following calculations now use the new attribute instead of the old one.

Some obsolete and/or unused code is updated/removed.


mbernasocchi:
LGTM

ptormene:
This PR addresses https://bugs.launchpad.net/oq-svir/+bug/1418610, together with https://github.com/gem/oq-svir-qgis/pull/18 and https://github.com/gem/oq-svir-qgis/pull/19

PR #19, opened 2015-02-12
Aggregate multivariate loss as with the saga workflow

ptormene:
When we switched to the use of SAGA for aggregating points by zone, the original workflow remained as a backup solution, but a big part of its code was duplicated and used to process the output produced by SAGA. In the old workflow, the geospatial search and the computation of zonal statistics (count of points, sum and average losses) were performed together. Although it might perform slightly better, the duplication was very annoying, because any necessary change to the computation of zonal statistics needed to be done twice. Furthermore, two not so straightforward tasks were performed together, reducing readability.
This PR changes the old workflow, making it consistent with the approach used by SAGA. Therefore, the aggregation now is done only to classify the points, labeling each one with the id of the zone to which it belongs. The statistics (count, sum and average loss) are computed afterwards, using the same code both in the case of using SAGA and in the case of using our backup geospatial search implementation.


mbernasocchi:
Hi @ptormene this also LGTM,

ptormene:
This PR contributes to address https://bugs.launchpad.net/oq-svir/+bug/1418610

PR #21, opened 2015-02-23
Check if two layers have the same length and the same contents

ptormene:
Add to ProcessLayer a method to check if a layer has the same length and content of another one. Some tests checking this functionality are added too.
The new method will be useful in a following PR, to check the results of 'loss aggregation by zone', comparing them with the expected results.


process_layer.py
```diff
@@ -44,6 +44,24 @@ class ProcessLayer():
     def __init__(self, layer):
         self.layer = layer
 
+    def has_same_content_as(self, other_layer):
+        this_dp = self.layer.dataProvider()
+        other_dp = other_layer.dataProvider()
+        len_this = sum(1 for feature in this_dp.getFeatures())
+        len_other = sum(1 for feature in other_dp.getFeatures())
+        if len_this != len_other:
```
mbernasocchi: why not simply:
self.layer.featureCount()  != other_layer.featureCount() :

mbernasocchi:
beside the one comment, LGTM

PR #22, opened 2015-02-24
Refactor loss aggregation

ptormene:
svir.py was becoming too big and it contained too many class variables. This PR moves all the code related to "loss aggregation by zone" into a separate module. Many class variables are substituted with a more proper use of function arguments. This allows to test each function separately without simulating the use of the GUI. Tests on loss aggregation will be added to this PR before merging it.


mbernasocchi:
LGTM

ptormene:
When running tests, SAGA is found and its version is detected, but the algorithm needed to label points by polygon is not found. We need to fix that before merging.

ptormene:
@mbernasocchi: are metadata_utilities.py and test_metadata_utilities.py needed or should we remove them from this repo?

mbernasocchi:
why would you want to get rid of them?? they are used in many places.

ptormene:
I noticed test_metadata_utilities.py is not run while running tests, so I thought it wasn't used

mbernasocchi:
LGTM

PR #23, opened 2015-02-26
Fix issue that we had when attempting to re-style the zonal layer after indices have been calculated

ptormene:
Avoid removing style (fill color) and style_border (solid line) and use default white and solid line.
Otherwise it's tricky for the user to restyle the layer after indices calculation, because the fill color and the border style need to be re-enable in order to make the polygons visible after a new classification.

### DIFF
--- a/svir.py
+++ b/svir.py
@@ -813,8 +813,9 @@ class Svir:
 
         not_null_rule = root_rule.children()[0].clone()
         not_null_rule.setSymbol(QgsFillSymbolV2.createSimple(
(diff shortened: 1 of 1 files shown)


ptormene:
This change has a drawback: the legend will display an additional square with black stroke and white fill. It might look a bit confusing. So we need to balance pros and cons and decide if we want to introduce this change or not.

ptormene:
This solution had a drawback and we re-implemented the fix in a different way

PR #24, opened 2015-02-27
Add a unique zone id to the zonal layer before aggregating losses using zonal geometries

ptormene:
NOTE: This PR is based on https://github.com/gem/oq-svir-qgis/pull/22, which should be merged first.

SAGA requires to specify one of the attributes of the zonal layer to be used as the id of the zone. In our zonal layer, though, we are not sure to have a field containing unique zone labels. This PR addresses this issue.
When aggregating loss by zone, if the user selects "use zonal geometries", then a unique id is assigned to each feature of the zonal layer, and it is used by SAGA to perform the "point in polygon" algorithm. If, instead, we already know in which zone each of the loss points is, we can select the field containing such information, and this is compared with one of the existing fields in the zonal layer specified by the user. All possible cases should be covered then. Tests have been changed, to work properly with such alternative workflows.


ptormene:
Closing and reopening towards master instead of towards refactor-aggregation

PR #25, opened 2015-03-03
When deleting attributes in case a calculation fails, use ProcessLayer

ptormene:
ProcessLayer.delete_attributes deletes attributes using LayerEditingManager, which starts each operation with layer.startEditing() and ends with layer.commitChanges() and layer.updateExtents().


ptormene:
Self-LGTM (very small fix)

PR #26, opened 2015-03-03
Allow negative weights for indicators

ptormene:
In case an indicator contributes "negatively" to building an index, this PR allows the user to set a negative weight to such indicator.
Instead of letting the _weights_ sum to 1, we let now the _absolute values of the weights_ summing to 1.
The size of tree nodes depends now on the absolute weight value too, as well as the positioning of the node labels.
When a negative value is assigned to a node, the stroke color of the node is swapped with the fill color, and these two colors are now bluish tones that should be easily recognized also by daltonic people. The circle's line stroke is now thicker, to make the border more visible (which is important when the circle is light and the fill is dark).
The code for all the calculations should remain as it is.
Such solution sounds very reasonable in case a weighted sum is performed. It might produce weird results when other operators are selected. We need to further investigate all the possible corner cases, before merging.


ptormene:
In case one of the operators that "ignore weights" is selected, the fill color of the nodes combined using such operator is set to "gold". In this way, the user is more clearly visually warned about the fact that those weights are going to be ignored.
I'm also displaying the weight values as percentages (approximated at the first decimal digit), to address a comment by the UME students. They were correctly suggesting that such information would be useful when printing the tree.
The changes produced by this PR can be seen in the screenshot.
screenshot from 2015-03-12 16 06 39

ptormene:
Current situation:
screenshot from 2015-03-18 09 31 27

resources/js/svir_viewer_d3_tree.js
```diff
@@ -666,6 +676,62 @@
                     updateButton(pdId);
                 });
 
+            // Write (ignore weights) in a new line, if present
```
bwyss: perhaps use 'render' in place of 'write', first time I look at this it was not so clear

resources/js/svir_viewer_d3_tree.js
```diff
@@ -666,6 +676,62 @@
...
+                .attr("x", function(d) { return Math.abs(d.weight) * CIRCLE_SCALE + 15; })
+                .attr("transform", "translate(0, 12)")
+                .style("fill", function(d) {
+                    if (d.operator !== undefined) {
+                        // Check for operators that ignore weights and style accordingly
+                        var color = '#660000';
+                        return color;
+                    }
+                })
+                .on("click", function(d) {
```
bwyss: Maybe this can be generalized into a function

resources/js/svir_viewer_d3_tree.js
```diff
@@ -95,9 +96,9 @@
             }
             $(function() {
                 $("#spinner-"+id).width(100).spinner({
-                    min: 0,
+                    min: -100,
                     max: 100,
-                    step: 0.01,
+                    step: 0.000001,
```
bwyss: I understand the reason for this, but the draw back is as a result the 'up' and 'down' arrows in the spinner are now useless

ptormene: I agree. I would like to collect opinions about this change of precision, from Marco and Chris, and then possibly change from spinner to text input.

resources/js/svir_viewer_d3_tree.js
ptormene: Agreed

resources/js/svir_viewer_d3_tree.js
```diff
@@ -666,6 +676,62 @@
...
+                    pdOperator = d.operator;
+                    pdId = d.id;
+                    $('#projectDefWeightDialog').empty();
+                    updated_nodes = d.children[0];
+                    findTreeBranchInfo(pdData, [pdName], [pdLevel], pdParentField, updated_nodes);
+                    operatorSelect(pdOperator);
+                    updateButton(pdId);
+                });
+
+            nodeEnter.append("text")
```
bwyss: Maybe add a small comment each time appending text to make note of what is being appended, now that there are many

resources/js/svir_viewer_d3_tree.js
ptormene: True

resources/js/svir_viewer_d3_tree.js
ptormene: Yes, I will add also some other comments and provide some simple examples to clarify what's happening

bwyss:
Besides the spinner issue, LGTM

mbernasocchi:
LGTM

resources/js/svir_viewer_d3_tree.js
```diff
@@ -32,6 +33,8 @@
             position: {my: "left top", at: "left top", of: "#projectDefDialog"},
             autoOpen: false,
             modal: true,
+            dialogClass: "no-close",
+            closeOnEscape: false,
```
bwyss: Why do you want to deactivate close on escape?

resources/js/svir_viewer_d3_tree.js
```diff
@@ -86,26 +89,34 @@
         function createSpinner(id, weight, name, field) {
             pdTempSpinnerIds.push("spinner-"+id);
             $('#projectDefWeightDialog').dialog("open");
-            if (field === undefined) {
-                $('#projectDefWeightDialog').append(
-                    '<p><label for="spinner'+id+'">'+name+': </label><input id="spinner-'+id+'" name="spinner" value="'+weight+'"></p>');
-            } else {
-                $('#projectDefWeightDialog').append(
-                    '<p><label for="spinner'+id+'">'+name+'('+field+'): </label><input id="spinner-'+id+'" name="spinner" value="'+weight+'"></p>');
+            var content = '<p><label style="clear: left; float: left; width: 10em;" for="spinner'+id+'">'+name;
```
bwyss: Consider moving the styling stuff into css

bwyss: should 'name' be followed by a colon ':' instead of a semi-colon ';' ?

ptormene:
I agree about the css. About the semi-colon, it's the end of the js line... It looks correct to me. Or are you meaning another `name`?

resources/js/svir_viewer_d3_tree.js
ptormene: The idea was to bind escape to a "cancel" function, more complex than just closing the dialog. I have to check, though, if we actually did it.

resources/js/svir_viewer_d3_tree.js
```diff
@@ -86,26 +89,34 @@
...
+            if (typeof field !== 'undefined') {
+                content += ' ('+field+')';
             }
+            content += ': </label><input id="spinner-'+id+'" name="spinner" value="'+Math.abs(weight)+'">';
+            content += '<input type="checkbox" id="inverter-spinner-'+id+'"><label style="font-size: 0.8em; "for="inverter-spinner-'+id+'" title="Select to invert the contribution of the variable to the calculation">Invert</label>';
```
bwyss: same as above with regards to the styling. 
Also consider line breaks like this:
'<input type="checkbox" id="inverter-spinner-'+id+'">' + 
'<label style="font-size: 0.8em; "for="inverter-spinner-'+id+'"' + 
'title="Select to invert the contribution of the variable to the calculation">Invert</label>';

resources/js/svir_viewer_d3_tree.js
```diff
@@ -256,24 +289,29 @@
...
 
                 // Adjust the values into percentages
                 pdTempWeights = pdTempWeights.map(Number);
                 var totalWeights = 0;
                 $.each(pdTempWeights,function() {
-                    totalWeights += this;
+                    totalWeights += Math.abs(this);
                 });
 
                 for (var i = 0; i < pdTempWeights.length; i++) {
```
bwyss: 'i' is already defined

resources/js/svir_viewer_d3_tree.js
```diff
@@ -256,24 +289,29 @@
...
-                    var tempMath = Math.floor((pdTempWeights[i] * 100) / totalWeights);
+                    var tempMath = (pdTempWeights[i] * 100) / totalWeights;
                     pdTempWeightsComputed.push(tempMath / 100);
                 }
 
                 // Update the results back into the spinners and to the d3.js chart
                 for (var i = 0; i < pdTempSpinnerIds.length; i++) {
```
bwyss: same as above 'i' is already defined

ptormene: Also my editor complains, but it sounds reasonable to re-define a counter, instead of using a different name in each loop.

resources/js/svir_viewer_d3_tree.js
ptormene: The semicolon ends the javascript instruction. It looks correct to me. But I agree there's something weird here. I don't see the `</p>`. I must check this.

PR #27, opened 2015-03-06
Force selecting the field containing the zone label, before uploading to the Platform

ptormene:
Before this change, after uploading a project to the Platform, a second user couldn't know which of the available layer's fields contains the zone labels used by the layer's creator. This PR forces the user to specify such field from a combobox, before proceeding with the upload. Such field gets saved in `project_definition['zone_label_field']`.
Instead of this approach, we could implement different solutions, but I'm not able to figure out any better one. For instance, we could save what zone id field was used when (and only if!) performing "loss aggregation by zone", and set such field as the default selection. In any case, the field might possibly be a numeric attribute, not so easily usable as a "label" to be displayed in Ben's charts in the web viewer. Therefore I believe it's reasonable to let the user assume the responsibility to decide, right before uploading.
NOTE: What needs to change on the web viewer, if this PR gets merged? I think, anyway, the user should be able to select a different field to be used to display labels on the charts, using the combobox that Ben has implemented (and that shouldn't be dismissed). The web viewer should use by default the `zone_label_field` specified in the project definition, unless the user explicitly selects a different field.

### DIFF
--- a/svir.py
+++ b/svir.py
@@ -1421,9 +1421,11 @@ class Svir:
         # convert bytes to MB
         file_size_mb = file_size_mb / 1024 / 1024
 
-        dlg = UploadSettingsDialog(file_size_mb)
+        dlg = UploadSettingsDialog(file_size_mb, self.iface)
         if dlg.exec_():
             project_definition['title'] = dlg.ui.title_le.text()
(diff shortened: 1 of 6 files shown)


mbernasocchi:
LGTM

PR #29, opened 2015-03-24
Fix corner case in which the project definition is defined in the qgis project, but it's empty

ptormene:
This is a one-line fix, but important. It solves a very annoying issue that might occur when the qgis project contains an empty project definition. In that case, the plugin attempts to parse the project definition's json, which is an empty string, so it raises an exception.
I added a check to avoid that.

### DIFF
--- a/svir.py
+++ b/svir.py
@@ -251,7 +251,7 @@ class Svir:
         # it returns a tuple, with the returned value and a boolean indicating
         # if such property is available
         resp = QgsProject.instance().readEntry('svir', 'project_definitions')
-        if resp[1] is True:
+        if resp[1] and resp[0]:
             self.project_definitions = json.loads(resp[0])
         else:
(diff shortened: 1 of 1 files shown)


svir.py
```diff
@@ -251,7 +251,7 @@ def sync_proj_def(self):
         # it returns a tuple, with the returned value and a boolean indicating
         # if such property is available
         resp = QgsProject.instance().readEntry('svir', 'project_definitions')
-        if resp[1] is True:
+        if resp[1] and resp[0]:
             self.project_definitions = json.loads(resp[0])
```
micheles: Can we have more telling names for resp[0] and resp[1]?

ptormene: Thanks Michele, I agree it was ugly

mbernasocchi:
LGTM

PR #30, opened 2015-03-26
Apply the QGIS style to the uploaded layer

ptormene:
This is a first attempt to style the geonode layer while uploading the QGIS project to the Platform.
Since the QGIS sld exporter is not compatible with geoserver's specifications, we had to re-implement the export. Currently this has been done just for the standard cases produced by the plugin. In the future it might be extended to enable applying more complex styles.


ptormene:
Closing this and opening a new PR towards master

PR #34, opened 2015-04-03
Fix some styling issues

ptormene:
Render the layer's style also while opening the weighting dialog (without waiting for any changes in the tree).
Save the layer's original style when the weighting dialog is opened, so it can be restored if "Cancel" is pressed in the dialog. 
Comment out lines that were resetting the fill to the style for null values. It did not work properly and it colored the zones with null values, instead of removing the fill.

### DIFF
--- a/svir.py
+++ b/svir.py
@@ -693,10 +693,8 @@ class Svir:
             self.update_actions_status()
         else:
             project_definition = old_project_definition
-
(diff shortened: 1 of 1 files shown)


mbernasocchi:
LGTM

mbernasocchi:
ah ups wip...

PR #35, opened 2015-04-07
Decouple the concept of 'variable inversion' from the weight's sign

ptormene:
Before this PR, we were using a hack, saving the 'invert variable' checkbox value into the variable's weight, by multiplying the weight by -1 in case of inversion, instead of saving the information as a separate variable. Such simplified implementation worked in most cases, but there was an annoying corner case. When using an operator that ignored weights, but still willing to 'invert' a variable, we were retrieving the sign of the weight, ignoring its absolute value. But unfortunately the sign was lost in case weight=0, because the effect of clicking the 'invert' checkbox was to multiply 0 by -1, obtaining -0, which resulted as 0.
This PR removes the hack, decoupling completely the concepts of 'weight' and 'inversion'. Each node of the weighting tree has now an additional 'isInverted' boolean attribute. This implementation is a bit more complicated than before, but it's logically cleaner and less confusing.


mbernasocchi:
LGTM

resources/js/svir_viewer_d3_tree.js
```diff
@@ -285,15 +285,14 @@
 
             function updateButtonClicked() {
                 pdTempWeights = [];
+                pdTempInverters = [];
                 pdTempWeightsComputed = [];
 
-                // Get the values of the spinners
+                // Get the values of the spinners and of the inverters
```
bwyss: could also be written "Get the values of the spinners and inverters"

resources/js/svir_viewer_d3_tree.js
```diff
@@ -338,7 +337,7 @@
             ))
             {
                 pdTempIds.push(pdData.id);
-                createSpinner(pdData.id, pdData.weight, pdData.name, pdData.field);
+                createSpinner(pdData.id, pdData.weight, pdData.name, pdData.field, pdData.isInverted);
```
bwyss: It is not clear to me where pdData.isInverted is defined?

ptormene: Nodes now contain an additional field called `isInverted`. A `invert` checkbox is used to populate such field. In this case, pdData is the node.

PR #37, opened 2015-04-09
Fix corner case where all sibling nodes have weight=0

ptormene:
There was a division by zero issue.


ptormene:
Chris gave his LGTM, so I am merging this

PR #38, opened 2015-04-15
Handle the case in which multiple project definitions are available for a single project

ptormene:
In that case, for now, we just use the most recent project definition.
In a later PR we will need to allow the user to select one of the available project definitions.
We will also need to have an additional workflow, to update one of the projects on the platform, adding only a new project definition to that (as the web viewer now does).

This PR also explicitly sets the parameters in json.dumps, in order to make the project definition format consistent in all cases, including the server-side occurrences.

### DIFF
--- a/metadata_utilities.py
+++ b/metadata_utilities.py
@@ -97,7 +97,10 @@ def generate_iso_metadata(project_definition=None):
         if DEBUG:
             print project_definition
         template_replacements['SVIR_PROJECT_DEFINITION'] = '<![CDATA[%s]]>' % \
-            json.dumps(project_definition)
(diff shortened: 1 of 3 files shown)


ptormene:
This is a quick and dirty solution, later improved by https://github.com/gem/oq-svir-qgis/pull/38, that is still WIP. It makes the plugin able to load projects from the Platform also in case multiple project definitions have been saved for a single layer using the web viewer. The main limitation of this PR is that the user will only be able to use the most recent project definition. The good thing is that this preliminary solution enables to test the loop with the web viewer, until the next PR will be properly completed.

mbernasocchi:
LGTM

PR #39, opened 2015-04-17
Allow selecting between the available project definitions

ptormene:
This PR improves https://github.com/gem/oq-svir-qgis/pull/38, that was a quick and dirty fix to be able to load projects from the platform after the web viewer adds new project definition versions to it.
The new changes allow the user to select one of the available project definitions and to save such selection into the QGIS project.  The change is not so trivial, because I had to change the structure of the object that collects the plugin's project data, and some of the logic. I've also added a new button to the plugin's toolkit and a corresponding new dialog to select one of the project definitions available for the active layer. The dialog contains a dropdown menu in which the items' labels are the titles of the project definitions. When one of the items is selected, a non-editable text area displays the whole selected project definition.


mbernasocchi:
LGTM

PR #41, opened 2015-04-21
Small fixes

ptormene:
* Fix a corner case in which the user attempts to set weights and there's no available project definition.
* Avoid printing the "parent" key into the project definition string


mbernasocchi:
LGTM

PR #43, opened 2015-04-22
Use QListWidget instead of QCombobox to list the available layers on the OQ-Platform

ptormene:
Otherwise, if the list is huge, it becomes very uncomfortable.
The GUI is also slightly improved to display the layer's info more clearly

### DIFF
--- a/download_layer_dialog.py
+++ b/download_layer_dialog.py
@@ -25,8 +25,9 @@
 """
 from xml.etree import ElementTree
 
+from PyQt4 import Qt
 from PyQt4.QtCore import pyqtSlot
-from PyQt4.QtGui import (QDialog, QDialogButtonBox)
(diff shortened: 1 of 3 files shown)


mbernasocchi:
```Traceback (most recent call last):
  File "/home/marco/.qgis2/python/plugins/oq-svir-qgis/svir.py", line 639, in import_layer
    self.update_proj_defs(layer.id(), project_definitions)
  File "/home/marco/.qgis2/python/plugins/oq-svir-qgis/svir.py", line 242, in update_proj_defs
    assert isinstance(proj_defs, list)
AssertionError```

ptormene:
https://github.com/gem/oq-svir-qgis/pull/44 should have solved that issue. I've merged that and I've aligned this branch with master, so now I think you shouldn't have the same error anymore.

mbernasocchi:
LGTM

PR #45, opened 2015-04-27
If the (stripped) hostname ends with '/', remove it

ptormene:
In the OQ-Platform connection settings, the user might set a host like `https://platform-dev.openquake.org/`, ending with `/`. Before this PR, this produced some weird behaviors, because the host was recognized, but the `/` interfered with the platform's API. In order to avoid this, we strip the hostname and remove the ending `/` if present.

### DIFF
--- a/settings_dialog.py
+++ b/settings_dialog.py
@@ -72,8 +72,11 @@ class SettingsDialog(QtGui.QDialog, Ui_SettingsDialog):
         Store the options into the user's stored session info.
         """
         mySettings = QtCore.QSettings()
-        mySettings.setValue('svir/platform_hostname',
-                            self.ui.hostnameEdit.text())
+        # if the hostname ends with '/', remove it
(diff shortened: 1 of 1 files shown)


settings_dialog.py
```diff
@@ -72,8 +72,11 @@ def saveState(self):
         Store the options into the user's stored session info.
         """
         mySettings = QtCore.QSettings()
-        mySettings.setValue('svir/platform_hostname',
-                            self.ui.hostnameEdit.text())
+        # if the hostname ends with '/', remove it
+        hostname = self.ui.hostnameEdit.text().strip()
+        if hostname.endswith('/'):
+            hostname = hostname[:-1]
```
micheles: You can also use .rstrip:

>>> 'pippo/'.rstrip('/')
'pippo'

This is better because it strips any number of slashes:

>>> 'pippo//'.rstrip('/')
'pippo'

micheles:
LGTM

PR #46, opened 2015-04-27
Stream the socioeconomic data downloaded from the platform and display a progress bar

ptormene:
This PR takes advantage of https://github.com/gem/oq-platform/pull/403 to stream the socioeconomic data downloaded from the Platform.
I am still having problems attempting to display the progress bar properly. Perhaps this will be solved as soon as the downloading will be performed in a separate thread.


ptormene:
@mbernasocchi: do you think we should we use WaitCursorManager also here?

mbernasocchi:
not really ast it will go in a thread

mbernasocchi:
LGTM

PR #47, opened 2015-04-28
Better manage platform login exceptions

ptormene:
There were some corner cases in which the login failed and the plugin broke instead of displaying an error message in the QGIS message bar. I also set a timeout of 10 seconds for the login, otherwise the server could not respond for a long time (also blocking the GUI).


mbernasocchi:
LGTM

PR #48, opened 2015-04-29
Recalculate indices after selecting a different project definition

ptormene:
Also avoid the plugin to break when the selected project definition is invalid (in such case, the indices are not recalculated and the style is not updated)


mbernasocchi:
LGTM

PR #49, opened 2015-04-30
Pre-populate the title field before uploading a project to the platform

ptormene:
In the "upload project to platform" dialog, if the project already has a title, use it instead of the default one.


mbernasocchi:
LGTM

PR #50, opened 2015-05-05
Enable to add a new project definition to a project already present on the platform

ptormene:
This PR completes the loop between the QGIS plugin and the web viewer.
When a project is loaded from the platform, the platform layer's id is saved into the project definitions. This information is retrieved when the user chooses to upload a project to the platform. In such situation, the user can decide to upload the current work as a new project (creating a new geonode layer) or to use the saved layer's id to update the corresponding project on the platform. The updating procedure uses the same API we made to allow the web viewer to append new project definitions into existing projects.

### DIFF
--- a/projects_manager_dialog.py
+++ b/projects_manager_dialog.py
@@ -71,11 +71,16 @@ class ProjectsManagerDialog(QDialog):
                                             'proj_defs': []}
         else:
             self.project_definitions = {'selected_idx': None, 'proj_defs': []}
-        for proj_def in self.project_definitions['proj_defs']:
-            if 'platform_layer_id' in proj_def:
-                self.platform_layer_id = proj_def['platform_layer_id']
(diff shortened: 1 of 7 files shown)


projects_manager_dialog.py
```diff
@@ -69,6 +71,11 @@ def get_project_definitions(self):
                                             'proj_defs': []}
         else:
             self.project_definitions = {'selected_idx': None, 'proj_defs': []}
+        for proj_def in self.project_definitions['proj_defs']:
+            if 'platform_layer_id' in proj_def:
+                self.platform_layer_id = proj_def['platform_layer_id']
+            if 'zone_label_field' in proj_def:
+                self.zone_label_field = proj_def['zone_label_field']
```
mbernasocchi: Why iterating and setting self.xxx?

mbernasocchi:
LGTM

PR #52, opened 2015-05-07
Simplify editing project definition title and description

ptormene:
This PR adds to the "Project definitions manager" a "Title" line edit and a "Description" text edit, that dynamically update the corresponding fields in the selected project definition.
NOTE: the project definition's "Description" is a brief summary of that specific project definition, and does NOT correspond to the "Abstract", which instead describes the whole layer and that is stored into the GeoNode layer's metadata.


mbernasocchi:
LGTM

PR #55, opened 2015-05-07
When setting the project definition's title in the project manager, also update the text in the combobox

ptormene:
This is a small fix


mbernasocchi:
Lgtm

PR #56, opened 2015-05-08
Apply style to the uploaded layer only if the style is rule-based (the only one we implemented yet)

ptormene:
Otherwise, simply use the default basic style, without attempting to convert the qgis style.
This situation can happen when we upload a layer that hasn't been styled by the plugin (for instance, when downloading some socioeconomic indicators and then uploading the layer without any further analysis). We might consider in the future if such situation should be allowed, or if we want to force the users to calculate some aggregated indices before uploading their project to the platform.


mbernasocchi:
LGTM

PR #57, opened 2015-05-11
Check layer projections before attempting to aggregate losses by zone

ptormene:
If the coordinate reference systems of the loss layer and of the zonal layer are different, avoid performing the aggregation and display an error message with the names of the projections.
Otherwise, the aggregation seemed to be performed without any warning/error, but with wrong and weird results.
In the future, we might want to consider the hypothesis of doing something similar to what ArcGIS does, i.e., assuming as a reference the projection of the first layer that is loaded, and using on the fly the same CRS for the other layers that are added later.


mbernasocchi:
LGTM

PR #60, opened 2015-05-20
Add "geometric mean" operator

ptormene:
The "geometric mean" operator is described in http://en.wikipedia.org/wiki/Geometric_mean
NOTE: We should implement the calculations for this operator also platform-side, otherwise projects created by the plugin will not be usable by the web viewer


mbernasocchi:
LGTM

ptormene:
We can merge it as soon as https://github.com/gem/oq-platform/pull/425 is ready

PR #61, opened 2015-05-21
In the d3 tree, let the primary indicators labels be on the right (and anchor text properly)

ptormene:
When the names of indicators (the final leaves of the tree) are long, we don't want them to be rendered at the left of their nodes, because it would cause them to overlap with other parts of the tree.


mbernasocchi:
LGTM

PR #63, opened 2015-06-05
Leverage the get_project_definitions API to get directly the supplemental_information…

ptormene:
…instead of doing 2 calls to get the metadata url and the metadata itself, and then parsing the metadata to isolate the supplemental_infomation.

The get_project_definition API method has been recently added to simplify the interface between the Integrated Risk Viewer and the Platform. With this PR, we obtain the same simplification also plugin-side.


mbernasocchi:
LGTM

PR #64, opened 2015-06-09
Refactor composite variable calculations and store themes as layer fields

ptormene:
This is an important improvement in the plugin, removing a lot of code duplication, improving modularity and enabling in the future to have a much more flexible project definition tree.
Furthermore, with this PR, intermediate composite variables (such as socioeconomic themes) are saved into new fields added to the layer. This is useful for data visualization, because it makes possible to style the layer classifying zones by the values of such fields (e.g., by "economy" or "population").


ptormene:
Implementing tests in https://github.com/gem/oq-svir-qgis/pull/65 I am fixing some minor issues of this PR, and the 2 PRs are tightly connected. Therefore I believe it would be simpler to close this one and review only https://github.com/gem/oq-svir-qgis/pull/65.

PR #65, opened 2015-06-19
Refactor the calculation of composite variables and add tests for it, for operators and inversions

ptormene:
This is an important improvement in the plugin, removing a lot of code duplication, improving modularity and enabling in the future to have a much more flexible project definition tree.
Furthermore, with this PR, intermediate composite variables (such as socioeconomic themes) are saved into new fields added to the layer. This is useful for data visualization, because it makes possible to style the layer classifying zones by the values of such fields (e.g., by "economy" or "population").

This PR also adds tests for the calculation of composite variables:
* Testing the `calculate_node` function on a single node, with all the available operators, inverting some variables
* Handling the corner case in which the geometric mean fails (root of negative values), collecting features with invalid data and allowing the user to toggle the selection of such features, with a separate button with respect to the one used to toggle the selection of features with missing values
* Testing the `calculate_composite_variable` function, checking that layer attributes are added/re-used correctly


mbernasocchi:
LGTM

PR #66, opened 2015-06-23
Add a button to the project manager, to make a copy of the selected project definition

ptormene:
Based on https://github.com/gem/oq-svir-qgis/pull/65

In addition to the new button for cloning a project definition, this PR fixes a bug in the form for updating the current project definition. Now, when the raw text is modified, the current project definition is updated, before it can be overwritten by changing the title or description.


mbernasocchi:
LGTM

PR #67, opened 2015-06-24
Add a dropdown menu to select the field to be used for styling the layer

ptormene:
Based on https://github.com/gem/oq-svir-qgis/pull/65

An additional dropdown menu is added to the weighting dialog, enabling the user to select which field should be used by the plugin to automatically style the layer. This makes possible, for instance, to quickly style the layer by one of the socioeconomic themes (that now are stored as fields in the layer).
If none is selected from this dropdown menu, the automatic behavior of the tool is the old one (check if it's possible to style by the IRI, otherwise attempt to style by the SVI, otherwise by the RI... or do nothing).

### DIFF
--- a/utils.py
+++ b/utils.py
@@ -89,6 +89,19 @@ def get_node(sub_tree, name):
     return None  # not found
 
 
+def get_field_names(sub_tree, field_names):
+    # return a list of all the field names defined in the project definition
+    # (browsing the tree recursively)
+    # field_names is an accumulator that needs to be passed as []
+    if 'field' in sub_tree:
+        field_names.append(sub_tree['field'])
+    if 'children' in sub_tree:
+        for child in sub_tree['children']:
+            child_field_names = get_field_names(child, field_names)
+            field_names.extend(child_field_names)
+    return field_names
+
+
 def get_credentials(iface):
     qs = QSettings()
     hostname = qs.value('svir/platform_hostname', '')
--- a/utils.py
+++ b/utils.py
@@ -89,10 +89,12 @@ def get_node(sub_tree, name):
     return None  # not found
 
 
-def get_field_names(sub_tree, field_names):
+def get_field_names(sub_tree, field_names=None):
     # return a list of all the field names defined in the project definition
-    # (browsing the tree recursively)
-    # field_names is an accumulator that needs to be passed as []
+    # field_names is an accumulator that is extended browsing the tree
+    # recursively
+    if field_names is None:
+        field_names = []
     if 'field' in sub_tree:
         field_names.append(sub_tree['field'])
     if 'children' in sub_tree:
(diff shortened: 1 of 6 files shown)


utils.py
```diff
@@ -89,6 +89,19 @@ def get_node(sub_tree, name):
     return None  # not found
 
 
+def get_field_names(sub_tree, field_names):
```
mbernasocchi: def get_field_names(sub_tree, field_names=None):

utils.py
```diff
@@ -89,6 +89,19 @@ def get_node(sub_tree, name):
...
+    # return a list of all the field names defined in the project definition
+    # (browsing the tree recursively)
+    # field_names is an accumulator that needs to be passed as []
```
mbernasocchi: if field_names is None:
    field_names = []

weight_data_dialog.py
```diff
@@ -89,12 +94,38 @@ def __init__(self, iface, project_definition):
...
         self.active_layer_numeric_fields = [
             field.name()
-            for field in iface.activeLayer().dataProvider().fields()
+            for field in self.iface.activeLayer().dataProvider().fields()
             if field.typeName() in NUMERIC_FIELD_TYPES]
 
+    def populate_style_by_field_cbx(self):
+        self.set_active_layer_numeric_fields()
+        # the empty list will be filled by get_field_names
+        fields_in_proj_def = get_field_names(self.project_definition, [])
```
mbernasocchi: fields_in_proj_def = get_field_names(self.project_definition)

weight_data_dialog.py
```diff
@@ -89,12 +94,38 @@ def __init__(self, iface, project_definition):
         self.frame.javaScriptWindowObjectCleared.connect(self.setup_js)
         self.web_view.loadFinished.connect(self.show_tree)
         self.json_updated.connect(self.handle_json_updated)
+        self.populate_style_by_field_cbx()
 
+    def update_project_definition(self, project_definition):
+        self.project_definition = deepcopy(project_definition)
+        self.populate_style_by_field_cbx()
+
+    def set_active_layer_numeric_fields(self):
```
mbernasocchi: a setter taking no value is a bit a strange thing

weight_data_dialog.py
```diff
@@ -89,12 +94,38 @@ def __init__(self, iface, project_definition):
...
+    def populate_style_by_field_cbx(self):
+        self.set_active_layer_numeric_fields()
+        # the empty list will be filled by get_field_names
+        fields_in_proj_def = get_field_names(self.project_definition, [])
+        fields_for_styling = [
+            field for field in self.active_layer_numeric_fields
+            if field in fields_in_proj_def]
+        # block signals to avoid performing the onchange actions while adding
+        # items programmatically
+        self.ui.style_by_field_cbx.blockSignals(True)
```
mbernasocchi: did you have a look at reload_attrib_cbx

ptormene: I can't reuse that, because I need to get only the fields that are both in the layer and in the project definition

weight_data_dialog.py
ptormene: Done

mbernasocchi:
Lgtm e arrivo

PR #68, opened 2015-06-26
Enable overwriting the transformed field or tracking the new field in the project definitions

ptormene:
When transforming a layer's field, it was not possible to overwrite the original field. This PR adds a checkbox enabling the user to save the result into the original field instead of creating a new one. The user is warned about the fact that this way the original values would be lost and that such operation can not be undone.
This change simplifies the main workflow very much, because it allows to transform the values of one of the nodes in the weighting tree, without losing its weight and extended name. Otherwise, normalizing a node and saving the new values into a different field, would make necessary to remove the initial node from the tree and add the new one, losing information.

In addition to that, I've improved the alternative workflow in which the user wants to keep the original attribute unchanged, saving the results of a transformation into a new field. In this case, there is a further additional checkbox that, if checked, lets the tool check if the original field was tracked by any of the project definitions, and links the project definitions to the new field name.


utils.py
```diff
@@ -45,6 +45,18 @@ def tr(message):
     return QApplication.translate('Svir', message)
 
 
+def replace_fields(sub_tree_root, before, after):
+    # recursively search the project definition for 'field's equal to the
```
mbernasocchi: not really clear doc

transformation_dialog.py
```diff
@@ -119,10 +137,16 @@ def on_algorithm_cbx_currentIndexChanged(self):
...
+    def on_attrib_cbx_currentIndexChanged(self):
+        self.update_default_fieldname()
+
     @pyqtSlot()
     def on_new_field_name_txt_editingFinished(self):
         self.attr_name_user_def = True
-        if not self.ui.new_field_name_txt.text():
+        input_field_name = self.ui.attrib_cbx.currentText()
+        new_field_name = self.ui.new_field_name_txt.text()
+        if not new_field_name or new_field_name == input_field_name:
```
mbernasocchi: comment why

ui/ui_transformation.ui
```diff
@@ -86,6 +92,39 @@
        </property>
       </widget>
      </item>
+     <item row="2" column="0">
+      <widget class="QCheckBox" name="overwrite_ckb">
+       <property name="text">
+        <string>Overwrite the field</string>
+       </property>
+       <property name="checked">
+        <bool>true</bool>
```
mbernasocchi: I think false would be a better default

PR #69, opened 2015-07-01
Fix a QGIS 2.10 issue with CSV import

ptormene:
QGIS 2.10 in unable to import multipolygons from CSV if there are whitespaces between polygons.
This quick fix removes those whitespaces before letting QGIS import the CSV.
Such change should be removed as soon as the bug will be fixed in QGIS.


mbernasocchi:
LeGiTtiMa

PR #70, opened 2015-07-01
To ensure compatibility with shapefiles, ensure that the names of new fields are max 10 chars long and do not contain spaces

ptormene:


mbernasocchi:
LGTM

PR #74, opened 2015-07-02
Show a confirmation/warning dialog when pressing "cancel" or "close" in the weighting dialog

ptormene:
This is to avoid the user to accidentally close the weighting dialog, losing all the unsaved changes. The same can be easily re-used for other dialogs, but it doesn't look necessary elsewhere (so far) and it might be a little annoying for the user to be asked to confirm everywhere, so it has been implemented only for the weighting dialog.


mbernasocchi:
LGTM

PR #76, opened 2015-07-02
Enable transforming multiple fields at once

ptormene:
If the user wants to apply the same transformation to many fields (e.g., normalizing all the indicators in a consistent way), it is very useful to be able to select multiple fields and perform the transformations with a single click, instead of repeating the same operation manually many times.


mbernasocchi:
Lgtm

PR #77, opened 2015-07-06
Minor improvements to the weighting dialog

ptormene:
* Improved editing of an indicator's name in the weighting dialog (using the label 'Name' instead of 'Description' and removing some redundancy between the debug html page and the real one, also solving the issue we had if enter was pressed instead of pressing "ok" after editing the node's name)
* Weighting dialog's elements expand correctly on window resize
* Improved multiselect widget


mbernasocchi:
LGTM

ui/list_multiselect_widget.py
```diff
@@ -141,6 +160,8 @@ def _setupUI(self):
         self.deselect_btn.setToolTip('Remove the selected items')
         self.select_all_btn.setToolTip('Add all')
         self.deselect_all_btn.setToolTip('Remove all')
+        spacer_label = QtGui.QLabel()
+        self.buttons_vertical_layout.addWidget(spacer_label)
```
mbernasocchi: why not using a spacer?

ptormene: It was a dirty way to make its height exactly the same as the "Unselected" and "Selected" labels, but I agree it would be cleaner to use a spacer. I am going to fix it.

PR #78, opened 2015-07-09
When aggregating loss by zone, let loss sum and average be NULL (instead of 0) for zones containing no loss points

ptormene:
Thanks to this change, when you calculate the integrated risk index, zones having a missing value (NULL) for risk will produce a NULL IRI. Otherwise, it would produce a potentially misleading result, considering zero losses where we actually don't have a measure of it.


mbernasocchi:
LGTM

PR #79, opened 2015-07-12
Fix function getRadius to render circles correctly also in case of ignoring weights

ptormene:
We had an issue in the rendering of nodes circles, that happened in some corner cases (not always) while changing the operator to someone ignoring weights.
I have tested this change and the problem seems to be fixed for all cases.


mbernasocchi:
LGTM

PR #80, opened 2015-07-12
Immediately style the layer by the selected field, even without closing the weighting dialog

ptormene:
We noticed that changing the field by which we wanted to style the layer, did not immediately produce a change to the style. Changes were displayed from the second time it was switched. I investigated, and I found out it was due to the fact we were not emitting a signal to declare the project definition was changed, so the styling was not performed right away.
Emitting the json_updated signal, the change becomes immediately visible, so it is possible to keep the weighting dialog open while styling by different fields.
It is a one-line change, but it produces an important improvement to the user-experience.


mbernasocchi:
LGTM

PR #81, opened 2015-07-12
Fix corner case in field transformation, updating project definition only if available

ptormene:
Otherwise an exception could happen (which was not catched properly)


mbernasocchi:
LGTM

PR #83, opened 2015-07-21
New supplemental information structure

ptormene:
Addressing https://bugs.launchpad.net/oq-svir/+bug/1463720
Changing the data structure of the supplemental information field, in order to keep separated the project definitions from other variables that are defined for the layer and that are shared by all the project definitions (e.g., the bounding box or the id of the layer on the Platform, that otherwise needed to be replicated for each project definition)


ptormene:
I've just spotted a problem in setting some project variables on updating a project that has just been uploaded to the platform (setting to WIP again)

ptormene:
I need to remove the selected_project_definition_idx from the supplemental information before it's uploaded to the platform, because we decided to keep it only locally

metadata_utilities.py
```diff
@@ -73,7 +73,7 @@ def _serialize_xml(write, elem, encoding, qnames, namespaces):
 # END MONKEYPATCH CDATA
 
 
-def generate_iso_metadata(project_definition=None):
+def generate_iso_metadata(supplemental_information=None):
```
bwyss: Just wondering about the terminology. If I understand this correctly, you are now calling a JSON object that contains N number of project definitions: 'supplemental information'? This name comes from the GeoNode layer field that is used to house the information, but perhaps we can consider other names that describe the JSON a bit better, like 'project definition container'? Or maybe the JSON is called a 'project definition' and then the blocks (that we are now calling  project definitions) are called something else, like 'project scheme'?

ptormene: What you suggest makes sense. Anyway, I think in this context supplemental_information is a good name, because I am building the metadata to be uploaded to geonode, where I am going to use the supplemental_information field. We could change the name inside QGIS, but I think it is intuitive to use the same name we are using on the platform, so the correspondence between the two things is more immediate. For now, I would prefer to keep the name as it is, but I am open to discuss further about it.

shared.py
```diff
@@ -119,6 +121,7 @@ def __repr__(self):
 
 
 PROJECT_TEMPLATE = {
+    'project_definition_version': '1.0',
```
mbernasocchi: use PROJECT_DEFINITION_VERSION

mbernasocchi:
beside the small comment LGTM

shared.py
ptormene: done

PR #85, opened 2015-08-05
Add latex version of the user manual

ptormene:
This PR adds a translation of the user manual into latex. I had to fix the basic structure, the formulas and the cross-references. The bibliography is currently managed as it was in the original text (using footnotes and inline citations, instead of putting everything in a separate bibliography). In the future this might be changed.
The styling is (intentionally) very basic and we will need to wait for a standard style to make this manual homogeneous with the other GEM user manuals.
I am keeping the manual inside the plugin's repository, but I am not 100% sure this is what we want (although it sounds quite reasonable to me).

### DIFF
--- a/manual/latex/.gitignore
+++ b/manual/latex/.gitignore
@@ -1,0 +1,5 @@
+*.aux
+*.pdf
+*.toc
+*.log
+*.out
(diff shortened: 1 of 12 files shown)


burtonchrisp:
I’m asking Vitor to send over theLaTex files used for the risk and hazard manuals. I’ll also send you James's cover page.

ptormene:
Thank you Chris, both will be useful

ptormene:
Missing headers with copyright etc.

ptormene:
We are trying a different approach: https://github.com/gem/oq-irmt-qgis/pull/98

PR #86, opened 2015-09-01
Show a warning dialog when attempting to add a node and no more numeric fields are available

ptormene:
Addressing https://bugs.launchpad.net/oq-svir/+bug/1490994

Side note: I've also renamed some counters, because my editor's syntax checker kept complaining about defining variables with the same name of already existing ones (even if they are in different scopes).


mbernasocchi:
I don't like your editor suggestion. beside that LGTM

ptormene:
I agree, it's quite annoying and I don't know how to disable it, so I decided to cut it off this way

PR #87, opened 2015-09-10
Count the active layer's vertices before uploading a project to the platform

ptormene:
We might want to set a maximum threshold for the complexity of geometries to be uploaded to the platform. With this PR, the vertices of the active layer's geometries are counted when the "upload" dialog is opened. This is still WIP, and we need to decide if and when we want to display the vertices count to the user, and if we want to forbid to upload something too complex or at least to display some kind of warning, also to inform the user that the web viewer will not be able to manage an exceedingly complex project.


ptormene:
I am replacing this PR with a similar one based on branch "new-supplemental-information-structure"

PR #88, opened 2015-09-11
Count the active layer's vertices before uploading a project to the platform

ptormene:
Based on https://github.com/gem/oq-svir-qgis/pull/83

We might want to set a maximum threshold for the complexity of geometries to be uploaded to the platform. With this PR, the vertices of the active layer's geometries are counted when the "upload" dialog is opened. This is still WIP, and we need to decide if and when we want to display the vertices count to the user, and if we want to forbid to upload something too complex or at least to display some kind of warning, also to inform the user that the web viewer will not be able to manage an exceedingly complex project.

### DIFF
--- a/process_layer.py
+++ b/process_layer.py
@@ -380,3 +380,49 @@ class ProcessLayer():
             return type_str in type_list
         else:
             return False
+
+    def count_vertices(self):
+        """
+        Inspired by "https://github.com/GeospatialEnablingTechnologies/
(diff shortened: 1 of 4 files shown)


mbernasocchi:
LGTM

ptormene:
Merging now, and postponing the decision about possible warnings or about limiting the complexity of geometries to be uploaded to the platform

PR #94, opened 2015-10-02
Rename plugin package back to svir due to QGIS constrains

mbernasocchi:
QGIS web plugin platform doesn't allow renaming the plugin package name so I refactored back to svir.
I also changed minimum QGIS version to 2.4 and bumped the plugin version to 1.7.1

*No other changes can go in master before this!*

after merging this, tag master as version 1.7.1


daniviga:
Ciao @mbernasocchi yes, I did some investigation yesterday and discovered the same. I'm doing some tests and then I'll give you the LGTM and merge.

daniviga:
LGTM, tested on 2.10 and 2.8.1, Linux and Windows

PR #95, opened 2015-10-29
Fix import error raised by QGIS 2.12 when importing QPyNullVariant

ptormene:
Importing `QPyNullVariant` from `qgis` doesn't work anymore.
It has to be imported from `PyQt4.QtCore` instead.


mbernasocchi:
see https://hub.qgis.org/issues/8258#journal-44357-notes
we could also use from qgis.core import NULL

but LGTM

PR #97, opened 2015-10-29
Fix style generation for QGIS version >= 2.12

ptormene:
Before QGIS version 2.12, `Rule` has the method `symbols()`,
which is replaced by `symbols2()` in QGIS version 2.12.
Also, before version 2.12, QgsSimpleLineSymbolLayerV2 always has the method `brushStyle`, but from version 2.12 this is not always true.
With this PR, the plugin should work with both QGIS versions.
Fixes https://github.com/gem/oq-irmt-qgis/issues/96


daniviga:
LGTM. Tested on Fedora 22 + QGIS 2.12 and worker without issues (new project, load project from platform). I would suggest you to mark current code as 1.7.2 and make a new release.

daniviga:
LGTM

daniviga:
LGTM

PR #98, opened 2015-11-04
Plugin's documentation produced via sphinx

ptormene:
This PR introduces some important improvements:
* the manual has been translated into rst files, to be compiled by sphinx, and it has been updated with respect to the recent changes in the plugin.
* sphinx is used also to produce a documentation of the code, from docstrings
* the manual and the code documentation contain cross-references. This is useful for instance to move from the section of the manual describing field transformations, to the part of the code where transformations are implemented and documented
* from the source code of the manual, sphinx also produces a pdf version, including a cover page that is consistent with GEM's CI.
* leveraging the "what's this?" functionality of QGIS, it is possible to click on each button of the plugin's toolbar, to be redirected to the page of the manual describing the button's functionality
* docstrings for the whole plugin have been reviewed and improved
* the plugin's makefile can now produce the versioned documentation and package it together with the plugin as well as uploading it to docs.openquake.org


mbernasocchi:
lgtm

PR #101, opened 2015-11-05
Keep selection of the 'style_layer_by' menu; fix #99

### LINKED ISSUE #99: In the weighting dialog, the selection of `style the layer by` is not kept after re-calculations
When indices are re-calculated, the plugin renders the map colors depending on the default field selection. Instead, if another node is selected, it should be re-used.

ptormene:


mbernasocchi:
Lgtm

PR #104, opened 2015-11-11
Don't break when OpenSSL has SSLv3_METHOD disabled.

mbernasocchi:
Don't break when OpenSSL has SSLv3_METHOD disabled.
this is needed at least for ubuntu 15.10

see http://stackoverflow.com/questions/28987891/#answer-29081240


ptormene:
LGTM

PR #106, opened 2015-11-11
Don't attempt to update project definitions after a transformation if no project definitions are available (fix #105)

### LINKED ISSUE #105: The plugin breaks in some corner cases when no project definition is defined
If, for a new project, the user loads a layer from their file system (instead of loading data from the OQ-Platform), still no project definition is defined. It breaks the plugin in workflows that require reading the available project definitions. Such problem has been encountered in 2 situations:
  1. transforming an attribute while the checkbox `Let the project definition's references track the new field(s)` is checked
  2. clicking the `Manage project definitions` button.

We need to properly handle the case in which no project definition is defined.

ptormene:


mbernasocchi:
lgtm

PR #108, opened 2015-11-20
Fix uploading on windows

ptormene:
For some strange reason, if the xml attached to the shapefile is opened in `'r'` mode instead of `'rb'`, the uploading does not work on windows (but it still works on ubuntu).
Changing it to `'rb'` make it works on both systems.

### DIFF
--- a/svir/metadata.txt
+++ b/svir/metadata.txt
@@ -25,6 +25,7 @@ email=staff.it@globalquakemodel.org
 changelog=
     1.7.4
     * Fix uploading issue on windows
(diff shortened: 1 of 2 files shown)


micheles:
The Python docs recommends to always use 'rb' for binary files, exactly because of the issue on Windows: https://docs.python.org/2/library/functions.html?highlight=open#open

daniviga:
LGTM, tested on Linux and Windows 7 (32bit)

ptormene:
We thought it was fine using `'r'` because the xml is not binary, but since `'rb'` doesn't break anything, we are ok with this change.

PR #109, opened 2015-11-24
Fix memory issue in weighting dialog (and more)

ptormene:
3 bugs fixed:
* We had a HUGE memory issue in the recursive procedure used to collect the field names in a project definition. It made the application exponentially slower the more the project definition became complex. It was causing the application to go out of memory even with relatively simple projects. Now that it is solved, things should be much faster and the weighting dialog shouldn't freeze even for much bigger and complex projects.
* When making the project definition track transformed fields, we had a problem in case one of the fields could not be transformed correctly (e.g. a corner case in which min-max transformation couldn't be performed because all the input values are equal to each other). In such case, the project definition was set to point to the transformed attribute, but the latter couldn't be found, later causing indices calculations to fail. With the proposed change, only fields that are successfully transformed will be tracked by the project definition.
* Fix https://github.com/gem/oq-irmt-qgis/issues/110

2 other changes:
* Automatic names for transformed fields will be `_` + `NAME` instead of `T_` + `NAME`. This saves one character, which for shapefiles can be very useful. Our indicator codes have 9 characters each, so the additional underscore is still within the 10 characters limit for field names in shapefiles. Therefore, our indicator names will not be truncated, whereas adding 2 characters would result in the truncation of the last character of the indicator's name.
* Display messages into the messageBar during field transformations and index calculations.

### DIFF
--- a/svir/calculations/calculate_utils.py
+++ b/svir/calculations/calculate_utils.py
@@ -196,7 +196,7 @@ def calculate_node(
         node, node_attr_name, node_attr_id, layer, discarded_feats):
     operator = node.get('operator', DEFAULT_OPERATOR)
     children = node['children']  # the existance of children should
-    # already be checked
+                                 # already be checked
     with LayerEditingManager(layer,
                              'Calculating %s' % node_attr_name,
                              DEBUG):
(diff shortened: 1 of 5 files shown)


mbernasocchi:
LGTM

PR #112, opened 2015-11-30
Shrink package

ptormene:
The QGIS repository does not accept plugins exceeding a size of 10MB, and the new documentation made the package become too big. We believe it is important to keep the documentation inside the package, so it remains available even working offline with no internet connection.
In order to keep the package as small as possible, I reduced the size of a png image and I removed from the package some folders that are not needed to run the plugin: tests, resources, source code for the documentation.


ptormene:
I've added to this PR also some corrections to the manual, following Chris's review.

mbernasocchi:
almost LGTM, /resources/icon.svg needs to be present in the package. You can either reinsert it or move it to /icon.svg and change metadata.txt to something like this icon=icon.svg

ptormene:
I have re-insterted resources in the package, because we are still almost 2MB below the threshold. We might consider to remove it again if we run out of space in the future.

PR #114, opened 2015-12-01
Fix get_irmt_version function

ptormene:
It was returning also an undesired newline after the version number, that broke the searching functionality in the documentation.


daniviga:
LGTM

PR #118, opened 2015-12-04
Synchronize correctly the d3 tree with changes made by on-the-fly recalculations...

### LINKED ISSUE #117: Wrong tree construction in a corner case when downloading indicators from Platform
When downloading indicators from the OQ-Platform, the project definition is automatically built using information downloaded from the socioeconomic database.
In case one of the indicators has exactly the same name as one the socioeconomic themes, the tree is built incorrectly. For instance, if one of the downloaded indicators is called 'Population', which is also the name of the theme to which it belongs, the 'Population' theme will be linked to the field that stores data for the primary indicator (instead of creating a new field for the theme).

ptormene:
…avoiding to make 2 nodes pointing to a same field, which happened in some corner cases (fix #117)


mbernasocchi:
LGTM

PR #122, opened 2015-12-10
Enable 'Project definitions manager' for new projects that are not yet associated to any project definition

ptormene:
Fix https://github.com/gem/oq-irmt-qgis/issues/121


mbernasocchi:
LGTM

PR #124, opened 2015-12-10
Enable zooming and panning in the d3 tree and avoid overlapping in the text labels

ptormene:
Fix https://github.com/gem/oq-irmt-qgis/issues/120

It seems to work correctly inside the plugin, but it causes some layout issues inside the `weight_data_debug.html` testing html page. Marco, please check about it.


mbernasocchi:
in QGIs it works nicely

mbernasocchi:
LGTM to me now

PR #125, opened 2015-12-14
Let the user set the shapefile name when downloading a project

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/119

Before the proposed change, when downloading a project from the OQ-Platform, the user selected just a destination folder. The names of the downloaded files were those taken from the zipfile produced server-side. With this change, the user sets both the destination folder and the file name, and the files extracted from the downloaded zipfile are renamed accordingly.


mbernasocchi:
LGTM

PR #126, opened 2015-12-14
Add to the weighting dialog a button to save the project definition tree as pdf

ptormene:
Fix https://github.com/gem/oq-irmt-qgis/issues/123

I first tried to add a button inside the html, linking it to a window.print() via javascript, but it looks like QWebView does not manage printing as inside a common web browser.
I changed the approach into adding a button to the dialog's GUI and assigning a QPrinter instance to the QWebView, explicitly assigning a specific output format.
When the printing button is pressed, the GUI asks for a destination file name through a separate dialog, and the printer is set to create the pdf in the specified folder and with the specified file name (with an additional check on possible overwriting).

The layout of the weighting dialog could be improved. Please, Marco, improve it a little bit if possible.


mbernasocchi:
lgtm, maybe the button would look better down near ok and cancel

PR #128, opened 2015-12-23
Use the QGIS logging system

ptormene:
Fix https://github.com/gem/oq-irmt-qgis/issues/127

Before this PR, all logging was done through simple `print` and QGIS wrote it to the console from which it was launched. It was causing problems on windows in case the python console was not instantiated. With this PR, the plugin leverages the QGIS logging system to write the logging messages into a IRMT section of the QGIS logging system.
Logging messages are printed to stderr when they are written from a test and the GUI is not available.
I've also improved some debug messages, pretty printing project definitions instead of displaying them in their raw representation.


svir/utilities/utils.py
```diff
@@ -56,6 +60,25 @@ def get_irmt_version():
...
+    :param level: the importance level
+        ('I' -> QgsMessageLog.INFO,
+         'W' -> QgsMessageLog.WARNING,
+         'C' -> QgsMessageLog.CRITICAL)
+    """
+    levels = {'I': QgsMessageLog.INFO,
+              'W': QgsMessageLog.WARNING,
+              'C': QgsMessageLog.CRITICAL}
+    if level not in levels.keys():
+        raise ValueError('Level must be one of %s' % levels.keys())
```
micheles: You can just write
```
if level not in levels:
```

micheles:
Apart the minor stylistic comment, LGTM

PR #131, opened 2016-02-17
Fix updating project title

ptormene:
suggesting to use the Project Definition Manager in case of duplication.

Only partially addressing https://github.com/gem/oq-irmt-qgis/issues/130


mbernasocchi:
lgtm

PR #134, opened 2016-03-14
Avoid removing border and fill from style

### LINKED ISSUE #129: The automatic styling sets 'No brush' and 'No pen', which can mislead the user
The automatic styling applied by the plugin sets `Fill style` to `No brush` and `Border style` to `No pen`. Therefore, if the user changes the style and forgets to re-set those attributes properly, all the features disappear from the map. We need to further investigate if this happens always or only in some corner cases.

We thought this issue was solved by https://github.com/gem/oq-irmt-qgis/pull/31
but this is happening again. Perhaps it could have been caused by recent changes in how QGIS handles styles.

ptormene:
Fix #129 

When automatically restyling a layer after a calculation, we were setting no-pen and no-brush in a part of the legend. If users wanted to apply a different style, they needed to explicitly reset the brush and pen style. In order to avoid this, we are not hiding that part of the legend anymore, and we are showing it with with a color that makes sense.

### DIFF
--- a/svir/irmt.py
+++ b/svir/irmt.py
@@ -915,8 +915,8 @@ class Irmt:
 
         not_null_rule = root_rule.children()[0].clone()
         not_null_rule.setSymbol(QgsFillSymbolV2.createSimple(
-            {'style': 'no',
-             'style_border': 'no'}))
+            {'color': '255,0,0,255',
+             'color_border': '0,0,0,255'}))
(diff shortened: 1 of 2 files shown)


ptormene:
This is a small change and it was done together with @mbernasocchi, so it's safe to merge it.

mbernasocchi:
Lgtm

PR #136, opened 2016-05-24
Change check on geometry type

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/135

`ProcessLayer.is_type_in` raises an exception if the layer's wkbType is unknown. This exception was not catched properly by the widget used to select layers for aggregating points in polygons. The undesired effect was that the presence of any "broken" layer in the list of available layers broke the tool, making it impossible to perform the aggregation even using "good" layers.
I fixed this wrong behavior by using the built-in `QgsVectorLayer.geometryType()` instead, which does not raise any exception in case the type is unknown.
NOTE: for some reason that is not so clear to me, QGIS has two different ways to classify the geometry type of a layer. One is by the enum QGis.GeometryType and one is by wkbType. In this case, the classification done through GeometryType (that is a bit more generic) looks sufficient.

### DIFF
--- a/svir/dialogs/select_input_layers_dialog.py
+++ b/svir/dialogs/select_input_layers_dialog.py
@@ -50,6 +50,16 @@ from svir.ui.ui_select_input_layers import Ui_SelectInputLayersDialog
 from svir.utilities.utils import tr, count_heading_commented_lines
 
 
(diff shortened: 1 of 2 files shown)


mbernasocchi:
I'll look into the redefinition. the rest looks good

mbernasocchi:
lgtm

PR #137, opened 2016-05-24
Add to the manual a section about using R scripts from QGIS

ptormene:
I added to the manual a slightly modified and re-adapted version of the blogpost http://blogs.openquake.org/devs/towards-driving-a-whole-integrated-risk-analysis-through-qgis.html#towards-driving-a-whole-integrated-risk-analysis-through-qgis


mbernasocchi:
lgtm

PR #142, opened 2016-10-14
Fix workflow using SAGA, in case of long field names

ptormene:
SAGA's clippointswithpolygons produces a shapefile with laundered names (reduced to max 10 characters). Therefore, the original field names can't be found in the output layer, unless we keep a map of old fieldnames -> laundered fieldnames.

In addition to this, I had to change the approach that was used to identify the field name for zone ids. Previously, we were assuming the loss_layer_plus_zone_ids would differ with respect to the loss_layer only by one additional field containing zone ids. But, as mentioned above, other field names could be laundered. So, the only thing we can do is to assume that the new field containing zone ids will be added "after" the existing ones, and to get the name of the last field.

Tests are green.

### DIFF
--- a/svir/calculations/aggregate_loss_by_zone.py
+++ b/svir/calculations/aggregate_loss_by_zone.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-#/***************************************************************************
+# /***************************************************************************
 # Irmt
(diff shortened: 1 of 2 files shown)


pslh:
Tested on Windows 10, works for me.
LGTM.

PR #146, opened 2016-11-04
Custom operator

ptormene:
This PR enables the user to include in the project definition also customized calculations that are not part of the set of operators defined in our plugin. When the "custom" operator is selected for a node, such node will not be recalculated by the tool, and the user will be able to select a layer's field where to store the results of customized calculations. It will also be possible to describe the custom calculation as free text. In the future, it would be better to leverage the QGIS field calculator at least for formulas that it supports.

There is also a platform-side companion PR (https://github.com/gem/oq-platform/pull/618), that enables the Integrated Risk Viewer (IRV) to visualize projects in which the concept of custom operator is used.


mbernasocchi:
lgtm

PR #157, opened 2017-01-27
Save/read the chosen directory and use it instead of the home directory

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/156


mbernasocchi:
LGTM

PR #158, opened 2017-01-27
Fix error handling while driving the oq-engine

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/155

This PR fixes the logic that checks the possible lockdown, that performs the login and that manages the life cycle of the dialog that drives the OQ Engine.


mbernasocchi:
LGTM

PR #161, opened 2017-01-31
Let community level curves save information only once

ptormene:
The community level recovery curve is a single curve for many selected points. It was inefficient to save the same information multiple times.


mbernasocchi:
LGTM

PR #162, opened 2017-01-31
Cleanup

ptormene:
Please see the descriptions of commits for a summary of changes.
The only user-visible addition is a new figure in the manual, showing an example of calculation log.


mbernasocchi:
LGTM

PR #163, opened 2017-02-07
Fix decimal separator issue in tests

ptormene:
Changes:
- revert commit 6fbfa5d6187a94ee34c418f849deb66f6361b807 ("Change test for custom operator")
- use a wrapper around every call to `QgsVectorFileWriter.writeAsVectorFormat` to set `LC_NUMERIC='C'`,  thus avoiding to use comma as decimal separator while writing shapefiles, in case the locale is, e.g., Italian)
- when writing the reference shapefiles for tests, use utf-8 encoding and the CRS of the initial layer, unless otherwise specified


mbernasocchi:
LGTM

PR #165, opened 2017-02-10
Customized automatic styling and possibility to avoid restyling at each recalculation

ptormene:
Still WIP, because we should re-use target fields when available, instead of creating new fields


ptormene:
This PR was replaced by https://github.com/gem/oq-irmt-qgis/pull/168 and https://github.com/gem/oq-irmt-qgis/pull/169

PR #169, opened 2017-02-14
Add a checkbox to enable/disable restyling the layer after each calculation

ptormene:
I decided to keep things much simpler, with respect to what I was trying to do in https://github.com/gem/oq-irmt-qgis/pull/165
A new checkbox is added to the plugin's settings, that is checked by default (keeping the usual behavior of the plugin). When it is unchecked, the plugin will not re-apply the default style after new recalculations of composite indices. This is useful whenever the user needs to apply a different kind of styling and he/she doesn't want the plugin to keep restyling the layer in a different way.
With the other approach, the setting was associated to each layer, so the user could force restyling some layers and not others. But the cost was a much more complicated mechanism, that required modifying the supplemental information structure and synchronizing that correctly between the main application and a couple of dialogs. Since I could not quickly fix some bugs I introduced while attempting to implement the other approach, I decided to implement this simpler version, which anyway should be enough in most use cases.

Addresses https://github.com/gem/oq-irmt-qgis/issues/148


ptormene:
We decided to save this settings at project level instead of at a general level. Therefore, the user will not have to remember to check it when switching between projects, and there will be a reduced risk of losing a customized styling.

mbernasocchi:
Good idea. I think it is a much better  approach.

ptormene:
I've implemented a mechanism that is similar to what I implemented to save/load recovery modeling configuration variables. If possible, the `force_restyling` option is associated to a specific layer; otherwise, it is associated to the project. The latest choice is also kept as a general setting. When reading the variable, the plugin first attempts to read it for the current layer, then for the current project, then from the general settings (finally picking the default if none of the above are available). I am not touching the supplemental information, so I am completely avoiding the synchronization issues that I had when I tried the other approach.

svir/utilities/utils.py
```diff
@@ -815,10 +816,28 @@ def get_style():
         'irmt/style_mode', style_mode_default, type=int)
     classes = settings.value(
         'irmt/style_classes', style_classes_default, type=int)
-
+    # look for the setting associated to the layer if available
+    force_restyling = None
+    if layer is not None:
+        # NOTE: I would use %s/%s instead of %s%s, but it does not work
+        value, found = QgsProject.instance().readEntry(
+            'irmt', '%s%s' % (layer.id(), 'force_restyling'))
```
mbernasocchi: why not %s_%s / is a specheal caracter in this case

svir/utilities/utils.py
```diff
@@ -815,10 +816,28 @@ def get_style():
...
+        if found:
+            force_restyling = True if value == 'True' else False
```
mbernasocchi: force_restyling = bool(value)

svir/dialogs/settings_dialog.py
```diff
@@ -138,13 +142,29 @@ def saveState(self):
...
 
         mySettings.setValue('irmt/style_mode', self.style_mode.itemData(
             self.style_mode.currentIndex()))
 
         mySettings.setValue('irmt/style_classes', self.style_classes.value())
+        active_layer = self.iface.activeLayer()
+        # at project level, save the setting associated to the layer if
+        # available
+        if active_layer is not None:
+            # NOTE: I would use %s/%s instead of %s%s, but it does not work
```
mbernasocchi: same note as below, why not _

svir/utilities/utils.py
ptormene: The value I read from QgsProject is a string, and bool(value) would be always True unless the string is empty. So I prefer to keep this unchanged.

PR #170, opened 2017-02-20
Fix adding features when importing data from oq-engine

ptormene:
Fix a bug that was introduced in https://github.com/gem/oq-irmt-qgis/pull/162
It was due to the fact that the methods named `addFeatures` in the QgsVectorLayer class and in the layer's dataProvider are slightly different (the former returns a boolean, and the latter returns a tuple; the former pre-selects features, the latter does not). I disabled the automatic selection, setting the makeSelected argument to False.


micheles:
LGTM, but it would be nice to have a test for this feature.

ptormene:
I agree it is important to test this part. I've added an issue (https://github.com/gem/oq-irmt-qgis/issues/172) and I will address that in another PR as soon as possible.

PR #173, opened 2017-02-20
Use readBoolEntry for QgsProject setting instead of the generic readEntry

ptormene:
As suggested by @mbernasocchi


mbernasocchi:
Fai i cowboy  merges? ;)

PR #180, opened 2017-02-23
Add a warning about the fact that imported layers are "in memory"...

ptormene:
...explaining how to save data and style persistently.


micheles:
LGTM

PR #184, opened 2017-03-01
Unify the management of exceptions while driving the oq-engine

ptormene:
The unification removes some duplication and improves readability. I am also handling the SSLError exception, that was not managed before and that stores the error message in a peculiar way.
I am also fixing the check on the connection, that has to be done whenever the dialog that drives the engine is created or opened.
Furthermore, I'm now checking both in cases in which the engine server's LOCKDOWN parameter is set to True or to False.

Fixes https://github.com/gem/oq-irmt-qgis/issues/183

### DIFF
--- a/svir/dialogs/drive_oq_engine_server_dialog.py
+++ b/svir/dialogs/drive_oq_engine_server_dialog.py
@@ -92,8 +92,8 @@ class DriveOqEngineServerDialog(QDialog, FORM_CLASS):
         try:
             self.login()
         except (ConnectionError, InvalidSchema, MissingSchema, ReadTimeout,
-                SvNetworkError):
-            self.is_logged_in = False
+                SvNetworkError, SSLError) as exc:
+            self._handle_generic_exception(exc)
         else:
             self.refresh_calc_list()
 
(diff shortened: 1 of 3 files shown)


svir/dialogs/drive_oq_engine_server_dialog.py
```diff
@@ -108,43 +114,25 @@ def login(self):
...
-                    self.is_logged_in = False
-                    raise
-                else:
-                    self.is_logged_in = True
-        if not self.is_logged_in:
-            msg = ("Please check OpenQuake Engine connection settings and"
-                   " credentials")
-            log_msg(msg, level='C', message_bar=self.iface.messageBar())
-            raise ConnectionError(msg)
+                # it can raise exceptions, catched by self.attempt_login
```
daniviga: catched -> cought

svir/dialogs/drive_oq_engine_server_dialog.py
```diff
@@ -108,43 +114,25 @@ def login(self):
...
-            try:
-                # FIXME: enable the user to set verify=True
-                resp = self.session.get(
-                    calc_list_url, timeout=10, verify=False)
-            except (ConnectionError, InvalidSchema, MissingSchema,
-                    ReadTimeout, SvNetworkError) as exc:
-                log_msg(str(exc), level='C',
-                        message_bar=self.iface.messageBar())
-                raise
+            # it can raise exceptions, catched by self.attempt_login
```
daniviga: catched -> caught

PR #191, opened 2017-03-03
Improve viewer layout

ptormene:
I made three main changes:

1. I found out how to remove all widgets from a layout, without having to remove them explicitly one by one
2. I picked a better size policy for all widgets so, for instance, the plotting area tends to occupy all the available space
3. I fixed https://github.com/gem/oq-irmt-qgis/issues/187, that was caused by an unnecessary call to adjustSize (the size is adjusted automatically anyway)

### DIFF
--- a/svir/dialogs/viewer_dock.py
+++ b/svir/dialogs/viewer_dock.py
@@ -215,13 +215,23 @@ class ViewerDock(QtGui.QDockWidget, FORM_CLASS):
         fill_fields_multiselect(
             self.fields_multiselect, self.iface.activeLayer())
 
-    def clear_layout(self, layout):
-        if not layout.count():
-            return
-        # a widget is deleted when it does not have a parent
+    def clear_widgets_from_layout(self, layout):
+        """
+        Recursively remove all widgets from the layout, except from nested
+        layouts. If any of such widgets is a layout, then clear its widgets
+        instead of deleting it.
+        """
         for i in reversed(range(layout.count())):
-            widget = layout.takeAt(i).widget()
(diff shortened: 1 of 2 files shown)


svir/dialogs/viewer_dock.py
```diff
@@ -241,25 +256,16 @@ def set_output_type_and_its_gui(self, new_output_type):
...
-            self.typeDepHLayout2)
-        self.remove_widgets_from_layout(
-            [self.warning_n_simulations_lbl,
-             self.fields_multiselect,
-             self.recalculate_curve_btn],
-            self.typeDepVLayout)
-        self.adjustSize()
+        # typeDepVLayout contains typeDepHLayout1 and typeDepHLayout2, that
+        # will be cleared recursively
+        self.clear_widgets_from_layout(self.typeDepVLayout)
```
micheles: Does it make sense to have a method with a single line?

svir/dialogs/viewer_dock.py
```diff
@@ -212,16 +214,29 @@ def create_fields_multiselect(self):
...
+        """
+        Recursively remove all widgets from the layout, except from nested
+        layouts. If any of such widgets is a layout, then clear its widgets
+        instead of deleting it.
+        """
+        for i in reversed(range(layout.count())):
+            item = layout.itemAt(i)
+            # check if the item is a nested layout
+            nested_layout = item.layout()
+            if nested_layout is not None:
```
micheles: I would call this `sublayout` instead of `nested_layout`

micheles:
LGTM

PR #195, opened 2017-03-10
Map selected fieldnames to laundered names

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/190

While aggregating points by zone, if the original layer contains long names, those names might be laundered to contain max 10 characters. When it happens, the selected original fields that contain damage state probabilities might not be found (not found returns -1), and the recovery modeling workflow would assume probabilities are contained in the field of index -1 (the last one) which instead contains the zone id. This was causing linux to produce bad results and windows to crash.
Here I am fixing it by mapping the original field names and the laundered ones, so the selected fields can actually be found and the calculation produces meaningful results.

While working on this I found out that a previous refactoring of some functions included in some functions' signatures and return values some unused objects, that now are removed.

### DIFF
--- a/svir/calculations/aggregate_loss_by_zone.py
+++ b/svir/calculations/aggregate_loss_by_zone.py
@@ -142,7 +142,7 @@ def calculate_zonal_stats(loss_layer,
                         zonal_layer.changeAttributeValue(
                                 feat.id(), unique_id_idx, feat.id())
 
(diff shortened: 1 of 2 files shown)


mbernasocchi:
LGTM

mbernasocchi:
LGTM

PR #197, opened 2017-03-13
Remove duplication of rlz in the layer names

ptormene:
Fix https://github.com/gem/oq-irmt-qgis/issues/192

Since the realization name is something like `rlz-XXX`, we don't need to add `rlz-`, otherwise it is duplicated.


mbernasocchi:
ora LGTM :)

PR #201, opened 2017-03-16
Add warnings

ptormene:
I am adding warnings in 2 cases:

1. asking confirmation before removing an OQ-Engine calculation (fixes https://github.com/gem/oq-irmt-qgis/issues/200)
2. notifying that a stored color setting was corrupted and the default value was restored instead (fixes https://github.com/gem/oq-irmt-qgis/issues/198)

### DIFF
--- a/svir/dialogs/drive_oq_engine_server_dialog.py
+++ b/svir/dialogs/drive_oq_engine_server_dialog.py
@@ -43,6 +43,7 @@ from PyQt4.QtGui import (QDialog,
                          QFileDialog,
                          QColor,
                          QDockWidget,
+                         QMessageBox,
                          )
 from svir.third_party.requests import Session
 from svir.third_party.requests.exceptions import (ConnectionError,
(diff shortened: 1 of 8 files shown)


svir/dialogs/drive_oq_engine_server_dialog.py
```diff
@@ -240,10 +241,16 @@ def on_calc_action_btn_clicked(self, calc_id, action):
...
+            confirmed = QMessageBox.question(
+                self,
+                'Remove calculation?',
+                'The calculation will be removed permanently. Are you sure?',
+                QMessageBox.Yes | QMessageBox.No)
+            if confirmed == QMessageBox.Yes:
+                self.remove_calc(calc_id)
+                if (self.current_output_calc_id is not None
+                        and self.current_output_calc_id == calc_id):
+                    self.clear_output_list()
```
micheles: I would just do

```
if self.current_output_calc_id == calc_id:
```

micheles:
LGTM

PR #202, opened 2017-03-17
Add a checkbox to load only the selected data from npz

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/193

I've also re-added the selection of the `poe` for `uhs`, because now we want to download data only for the chosen poe.

### DIFF
--- a/svir/test/test_load_npz_as_layer.py
+++ b/svir/test/test_load_npz_as_layer.py
@@ -87,6 +87,7 @@ class LoadNpzAsLayerTestCase(unittest.TestCase):
     def test_load_uhs(self):
         filepath = os.path.join(self.data_dir_name, 'output-184-uhs_67.npz')
         dlg = LoadNpzAsLayerDialog(IFACE, 'uhs', filepath)
+        dlg.load_selected_only_ckb.setChecked(False)
         dlg.accept()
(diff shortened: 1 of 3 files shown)


svir/test/test_load_npz_as_layer.py
```diff
@@ -84,9 +84,23 @@ def test_load_hazard_curves(self):
...
+        # test exporting the current selection to csv
+        self._test_export('uniform_hazard_spectra.csv')
+
+    def test_load_uhs_all(self):
+        filepath = os.path.join(self.data_dir_name, 'output-184-uhs_67.npz')
+        dlg = LoadNpzAsLayerDialog(IFACE, 'uhs', filepath)
+        dlg.load_selected_only_ckb.setChecked(False)
         dlg.accept()
         # FIXME: setActiveLayer is not working. As a workaround, I am deleting
         # all layers except the one I am testing.
```
micheles: It the FIXME comment still valid?

micheles:
LGTM

svir/test/test_load_npz_as_layer.py
ptormene: Yes, the comment is still valid. See also https://github.com/gem/oq-irmt-qgis/issues/177

PR #203, opened 2017-03-17
Display the data shape for the chosen rlz

ptormene:


micheles:
Instead of an opaque `shape: (528,)` it would be better to display something more significant, like `num_sites: 528` since those are one-dimensional arrays.

svir/dialogs/load_npz_as_layer_dialog.py
```diff
@@ -365,6 +362,13 @@ def populate_rlz_cbx(self):
         # self.rlz_cbx.addItem('All')
         self.rlz_cbx.addItems(self.rlzs)
 
+    def show_num_sites(self):
+        # NOTE: we are assuming all realizations have the same number of sites.
+        #       If different realizations have a different number of sites, we
+        #       need to move this block of code inside on_rlz_changed()
```
micheles: The assumption is always correct

svir/dialogs/load_npz_as_layer_dialog.py
```diff
@@ -109,6 +111,11 @@ def define_gui_elements(self):
         self.rlz_cbx.setEnabled(False)
         self.rlz_cbx.currentIndexChanged['QString'].connect(
             self.on_rlz_changed)
+        self.shape_msg = 'Number of sites: %s'
+        self.rlz_shape_lbl = QLabel(self.shape_msg % '')
```
micheles: Rather than `shape_msg` you could call it `num_sites_msg`.

micheles:
Apart the minor comments, LGTM

PR #208, opened 2017-03-28
Single dialog to load all oq-engine outputs as layers, plus loader for ruptures

ptormene:
I am introducing the following improvements:

1. using a single dialog to load all kinds of oq-engine outputs as layer, that adds to itself the right widgets with respect to the output type selected through ad dropdown menu (it required a lot of refactoring and many file names were changed, so unfortunately the review process might be tricky)
2. ruptures can be loaded automatically
3. I've added tests for loading ruptures and dmg_by_asset
4. I've removed the functionality loading oq-engine outputs from geojson, that were replaced by npz


svir/dialogs/load_output_as_layer_dialog.py
```diff
@@ -53,58 +57,80 @@
...
+                        self.npz_file = numpy.load(self.path, 'r')
+                        self.populate_out_dep_widgets()
+            self.file_browser_tbn.setEnabled(True)
+        else:
+            self.file_browser_tbn.setEnabled(False)
+        if self.path:
+            if output_type == 'dmg_by_asset':
+                self.read_loss_types_and_dmg_states_from_csv_header()
+            elif self.output_type != 'ruptures':
+                # we load everything
```
micheles: The comment is unclear since there is a pass below

svir/dialogs/load_output_as_layer_dialog.py
```diff
@@ -53,58 +57,80 @@
...
+                pass
+
+    def populate_output_type_cbx(self):
+        self.output_type_cbx.clear()
+        self.output_type_cbx.addItem('')
+        self.output_type_cbx.addItems(list(OQ_ALL_LOADABLE_TYPES))
```
micheles: I would use `sorted(OQ_ALL_LOADABLE_TYPES)` here

micheles:
There are too many methods for  a single widget which is doing everything. You will need to refactor this later on, when other importers will be implemented. But for the moment LGTM.

PR #210, opened 2017-03-29
Show full report

ptormene:
Based on https://github.com/gem/oq-irmt-qgis/pull/208

This PR adds the possibility to visualize the full report of a calculation directly inside a scrollable widget, instead of downloading it and looking for the file from an external tool.


micheles:
Non ASCII characters are not displayed correctly, but they are still visibile and nothing is broken too badly. LGTM.

PR #211, opened 2017-03-30
Make possible to run a calculation on top of an existing risk calculation

ptormene:
The Web UI should have the same behavior.


ptormene:
Originally the web API was designed to run risk calculations on top of an existing hazard calculation, so it requires the parameter `hazard_job_id`. That parameter should be renamed into `base_job_id` or something like that.

svir/dialogs/drive_oq_engine_server_dialog.py
```diff
@@ -199,16 +199,9 @@ def refresh_calc_list(self):
...
-                    if (calc['job_type'] == 'risk'
-                            or calc['status'] != 'complete'):
-                        continue
-                # do not display the button for outputs until calc is complete
-                elif action['label'] == 'Outputs':
-                    if calc['status'] != 'complete':
-                        continue
+                if (calc['status'] != 'complete' and
+                        action['label'] in ('Continue', 'Outputs')):
+                    continue
```
micheles: Perhaps a comment like "display the Continue or Output buttons only if the computation is completed"

micheles:
LGTM

PR #212, opened 2017-04-04
Update the test files for loading ruptures from csv

ptormene:
The ruptures.csv file produced by the OQ-Engine has changed its format in https://github.com/gem/oq-engine/pull/2694
The behavior of the plugin was not affected by that change, but I am updating the test files to have the same format produced by the engine.
This situation highlights the need for integration tests.


micheles:
Blocked by https://github.com/gem/oq-engine/issues/2673

micheles:
This is unblocked now. Here is an example of the CSV format for the ruptures: https://github.com/gem/oq-engine/blob/master/openquake/qa_tests_data/event_based/case_5/expected/ruptures.csv

PR #213, opened 2017-04-07
New set of icons and actions added optionally to the toolbar

ptormene:
The toolbar was removed because it was becoming too crowded. I re-added it, with an additional parameter to decide if an action has to be also added to the toolbar or only to the menu.
Some icons were previously missing at all; others have been replaced. In some cases, the new version is just the old one embedded in a OQ symbol, that represents the new theme for all the icons.
I've also updated the manual accordingly.


mbernasocchi:
obviously before merging we need to update metadata.txt

PR #215, opened 2017-04-12
Loaders of oq-engine outputs refactored using inheritance

ptormene:
Before this big refactoring, a single class was used to load every possible kinds of oq-engine outputs for which we have already implemented such functionality. Keeping everything in one single place is ok for a small number of output types, but it is not easily scalable. The aim of this refactoring was to identify the portions of code that all loaders have in common and those that need to be implemented differently for each output type. A main class takes care of the common parts and it defines some methods that are performed in the same by most loaders. For each output type I have defined a subclass that inherits the common behaviors and that overrides the methods that needed to be re-implemented specifically.

Although the total number of lines of code has increased, many lines are just comments (e.g. copyright strings), imports, additional menu items, etc. Furthermore, the file that was becoming problematically big (svir/dialogs/load_output_as_layer_dialog.py) has been shortened a lot, and it will be reduced much more as soon as drafts for other loaders (risk stuff that was implemented for hdf5 and that need to be rebuilt) will be moved from there into their own classes. Thanks to the current structure, the addition of new loaders will be much simpler.

### DIFF
--- a/svir/dialogs/load_dmg_by_asset_as_layer_dialog.py
+++ b/svir/dialogs/load_dmg_by_asset_as_layer_dialog.py
@@ -1,0 +1,113 @@
+# -*- coding: utf-8 -*-
+# /***************************************************************************
+# Irmt
+#                                 A QGIS plugin
+# OpenQuake Integrated Risk Modelling Toolkit
+#                              -------------------
+#        begin                : 2013-10-24
+#        copyright            : (C) 2014 by GEM Foundation
+#        email                : devops@openquake.org
+# ***************************************************************************/
+#
+# OpenQuake is free software: you can redistribute it and/or modify it
+# under the terms of the GNU Affero General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# OpenQuake is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with OpenQuake.  If not, see <http://www.gnu.org/licenses/>.
+
+import csv
+import tempfile
+from PyQt4.QtCore import pyqtSlot
+from svir.utilities.utils import import_layer_from_csv
+from svir.dialogs.load_output_as_layer_dialog import LoadOutputAsLayerDialog
+
+
+class LoadDmgByAssetAsLayerDialog(LoadOutputAsLayerDialog):
+    """
+    Modal dialog to load ruptures from an oq-engine output, as layer
+    """
+
+    def __init__(self, iface, output_type='ruptures', path=None, mode=None):
+        LoadOutputAsLayerDialog.__init__(self, iface, output_type, path, mode)
+        self.ok_button.setDisabled(True)
+        self.file_browser_tbn.setEnabled(False)
+        self.create_dmg_state_selector()
+        self.create_loss_type_selector()
+        self.create_save_as_shp_ckb()
+        self.setWindowTitle('Load scenario damage by asset from CSV, as layer')
+        self.adjustSize()
+        self.set_ok_button()
+        if self.path:
+            self.read_loss_types_and_dmg_states_from_csv_header()
+
+    @pyqtSlot()
+    def on_file_browser_tbn_clicked(self):
+        if self.open_file_dialog():
+            # read the header of the csv, so we can select from its fields
+            self.read_loss_types_and_dmg_states_from_csv_header()
+
+    def set_ok_button(self):
+        self.ok_button.setEnabled(
+            bool(self.path)
+            and self.dmg_state_cbx.currentIndex() != -1
+            and self.loss_type_cbx.currentIndex() != -1)
+        self.ok_button.setEnabled(bool(self.path))
+
+    def load_from_csv(self):
+        if self.mode == 'testing':
+            dest_shp = tempfile.mkstemp(suffix='.shp')[1]
+        else:
+            dest_shp = None  # the destination file will be selected via GUI
+        self.layer = import_layer_from_csv(
+            self, self.path_le.text(), self.output_type,
+            self.iface.messageBar(),
+            save_as_shp=self.save_as_shp_ckb.isChecked(), dest_shp=dest_shp)
+        dmg_state = self.dmg_state_cbx.currentText()
+        loss_type = self.loss_type_cbx.currentText()
+        field_idx = -1  # default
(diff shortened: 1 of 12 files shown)


svir/dialogs/load_output_as_layer_dialog.py
```diff
@@ -85,50 +68,26 @@ class LoadOutputAsLayerDialog(QDialog, FORM_CLASS):
...
         # sanity check
-        if (output_type is not None
-                and output_type not in OQ_ALL_LOADABLE_TYPES):
+        if output_type not in OQ_ALL_LOADABLE_TYPES:
             raise NotImplementedError(output_type)
         self.iface = iface
         self.path = path
-        self.npz_file = None
+        # self.npz_file = None
         self.output_type = output_type
```
micheles: Should this be commented or not? Usually one sets the attribute to None with a comment that the value is set in subclasses.

svir/dialogs/load_output_as_layer_dialog.py
```diff
@@ -162,7 +121,7 @@ def create_poe_selector(self):
 
     def create_loss_type_selector(self):
         self.loss_type_lbl = QLabel(
-            'Loss Type (used for default styling)')
+            'Loss Type')
         self.loss_type_cbx = QComboBox()
```
micheles: No need to use two lines

svir/dialogs/load_ruptures_as_layer_dialog.py
```diff
@@ -0,0 +1,60 @@
...
+from svir.dialogs.load_output_as_layer_dialog import LoadOutputAsLayerDialog
+
+
+class LoadRupturesAsLayerDialog(LoadOutputAsLayerDialog):
+    """
+    Modal dialog to load ruptures from an oq-engine output, as layer
+    """
+
+    def __init__(self, iface, output_type='ruptures', path=None, mode=None):
+        assert(output_type == 'ruptures')
```
micheles: Parens unneeded

svir/dialogs/load_uhs_as_layer_dialog.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class LoadUhsAsLayerDialog(LoadOutputAsLayerDialog):
+    """
+    Modal dialog to load uniform hazard spectra from an oq-engine output,
+    as layer
+    """
+
+    def __init__(self, iface, output_type='uhs', path=None, mode=None):
+        assert(output_type == 'uhs')
```
micheles: Parens unneeded

svir/dialogs/load_dmg_by_asset_as_layer_dialog.py
```diff
@@ -0,0 +1,110 @@
...
+            self, self.path_le.text(), self.output_type,
+            self.iface.messageBar(),
+            save_as_shp=self.save_as_shp_ckb.isChecked(), dest_shp=dest_shp,
+            zoom_to_layer=zoom_to_layer)
+        dmg_state = self.dmg_state_cbx.currentText()
+        loss_type = self.loss_type_cbx.currentText()
+        field_idx = -1  # default
+        for idx, name in enumerate(self.csv_header):
+            if dmg_state in name and loss_type in name and 'mean' in name:
+                field_idx = idx
```
ptormene: Yes, I agree it would be safer to raise an error instead

svir/dialogs/load_output_as_layer_dialog.py
ptormene: I prefer to avoid having the attribute `npz_file` in all subclasses, because some of them do not load data from npz, but from csv. So I am deleting it.

svir/dialogs/load_gmf_data_as_layer_dialog.py
```diff
@@ -0,0 +1,140 @@
...
+    def build_layer_name(self, rlz, **kwargs):
+        self.imt = self.imt_cbx.currentText()
+        self.eid = self.eid_sbx.value()
+        self.default_field_name = '%s-%s' % (self.imt, self.eid)
+        # layer_name = "gmf_data_%s_eid-%s" % (rlz, self.eid)
+        layer_name = "scenario_damage_gmfs_%s_eid-%s" % (rlz, self.eid)
+        return layer_name
+
+    def get_field_names(self, **kwargs):
+        field_names = list(self.dataset.dtype.names)
```
ptormene: Yes, I need a list instead of a tuple, because afterwards I need to tweak that list, while adding fields to the layer I am creating. It is because I have to keep track of how field names are laundered (for instance to be compliant with shapefiles constraints).

svir/dialogs/load_dmg_by_asset_as_layer_dialog.py
```diff
@@ -0,0 +1,110 @@
...
+    def read_loss_types_and_dmg_states_from_csv_header(self):
+        with open(self.path, "rb") as source:
+            reader = csv.reader(source)
+            self.csv_header = reader.next()
+            # ignore asset_ref, taxonomy, lon, lat
+            names = self.csv_header[4:]
+            # extract from column names such as: structural~no_damage_mean
+            loss_types = set([name.split('~')[0] for name in names])
+            dmg_states = set(['_'.join(name.split('~')[1].split('_')[:-1])
+                              for name in names])
```
ptormene: I prefer to keep this here, since it is not used anywhere else yet. If new loaders will require the same, I will move it to the external utilities. I will try to make it clearer, though.

svir/dialogs/load_gmf_data_as_layer_dialog.py
micheles: Please add a comment somewhere about this logic

svir/dialogs/load_dmg_by_asset_as_layer_dialog.py
```diff
@@ -90,12 +90,21 @@ def read_loss_types_and_dmg_states_from_csv_header(self):
...
+            loss_types = set()
+            dmg_states = set()
+            for name in names:
+                # each name looks like: structural~no_damage_mean
+                loss_type, dmg_state_plus_stat = name.split('~')
+                # dmg_state_plus_stat looks like: no_damage_mean
+                dmg_state, _ = dmg_state_plus_stat.rsplit('_', 1)
+                loss_types.add(loss_type)
+                dmg_states.add(dmg_state)
             self.populate_loss_type_cbx(list(loss_types))
```
micheles: I would use sorted here instead of set, just to make sure that there is a fixed order

micheles: And the same the line below

micheles:
Apart the last minute comments, LGTM

svir/dialogs/load_gmf_data_as_layer_dialog.py
ptormene: Done

PR #220, opened 2017-05-04
Load stats as layers

ptormene:
When a OQ Engine calculation has more than one realization, the corresponding output does not contain results for any of the realizations, but it contains statistical outputs (e.g. mean, quantiles). With the same widget that was used to select one of the available realizations, now it is possible to select the single existing realization or one of the statistical outputs.
Since the method `populate_rlz_cbx` does the same for many loaders, I don't need to override that method in the subclasses anymore.

I've also fixed a little bug in the creation of the layer name for hazard curves.

Fixes https://github.com/gem/oq-irmt-qgis/issues/219

### DIFF
--- a/svir/dialogs/load_output_as_layer_dialog.py
+++ b/svir/dialogs/load_output_as_layer_dialog.py
@@ -86,7 +86,6 @@ class LoadOutputAsLayerDialog(QDialog, FORM_CLASS):
         clear_widgets_from_layout(self.output_dep_vlayout)
 
     def create_rlz_selector(self):
-        self.rlz_lbl = QLabel('Realization')
         self.rlz_cbx = QComboBox()
         self.rlz_cbx.setEnabled(False)
         self.rlz_cbx.currentIndexChanged['QString'].connect(
@@ -272,9 +271,9 @@ class LoadOutputAsLayerDialog(QDialog, FORM_CLASS):
         #         self.hdata = self.npz_file['loss_maps-rlzs']
         #     _, n_rlzs = self.hdata.shape
         #     self.rlzs = [str(i+1) for i in range(n_rlzs)]
+        self.rlzs = [key for key in self.npz_file.keys() if key != 'imtls']
         self.rlz_cbx.clear()
         self.rlz_cbx.setEnabled(True)
-        # self.rlz_cbx.addItem('All')
         self.rlz_cbx.addItems(self.rlzs)
 
     def populate_dmg_state_cbx(self, dmg_states):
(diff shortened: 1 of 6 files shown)


svir/dialogs/load_output_as_layer_dialog.py
```diff
@@ -272,9 +271,9 @@ def populate_rlz_cbx(self):
         #         self.hdata = self.npz_file['loss_maps-rlzs']
         #     _, n_rlzs = self.hdata.shape
         #     self.rlzs = [str(i+1) for i in range(n_rlzs)]
+        self.rlzs = [key for key in self.npz_file.keys() if key != 'imtls']
         self.rlz_cbx.clear()
```
micheles: I am not sure if the call to .keys() is needed. I would replace it with a `sorted(self.npz_file)`.

svir/dialogs/load_output_as_layer_dialog.py
```diff
@@ -307,10 +306,9 @@ def get_layer_group(self, rlz):
         # get the root of layerTree, in order to add groups of layers
         # (one group for each realization)
         root = QgsProject.instance().layerTreeRoot()
-        group_name = 'Realization %s' % rlz
-        rlz_group = root.findGroup(group_name)
+        rlz_group = root.findGroup(rlz)
         if not rlz_group:
```
micheles: I would use a more generic name.  Instead of `rlz` something like `npz_key`.

micheles:
Apart the minor comments, LGTM

micheles:
It does not work on Python 3 with an error:
```
2017-05-04T15:09:35	1	Traceback (most recent call last):
			  File "/home/michele/.qgis2/python/plugins/svir/dialogs/drive_oq_engine_server_dialog.py", line 481, in <lambda>
			    self.on_output_action_btn_clicked(output, action, outtype))
			  File "/home/michele/.qgis2/python/plugins/svir/dialogs/drive_oq_engine_server_dialog.py", line 517, in on_output_action_btn_clicked
			    self.iface, output_type, filepath)
			  File "/home/michele/.qgis2/python/plugins/svir/dialogs/load_hmaps_as_layer_dialog.py", line 52, in __init__
			    self.populate_out_dep_widgets()
			  File "/home/michele/.qgis2/python/plugins/svir/dialogs/load_output_as_layer_dialog.py", line 255, in populate_out_dep_widgets
			    self.show_num_sites()
			  File "/home/michele/.qgis2/python/plugins/svir/dialogs/load_output_as_layer_dialog.py", line 295, in show_num_sites
			    rlz_data = self.npz_file[self.rlz_cbx.currentText()]
			  File "/usr/lib/python2.7/dist-packages/numpy/lib/npyio.py", line 228, in __getitem__
			    raise KeyError("%s is not a file in the archive" % key)
			KeyError: u' is not a file in the archive'
```

micheles:
LGTM

PR #223, opened 2017-05-04
Fix behavior of 'cancel' button in the dialogs that load OQ Engine outputs

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/222


micheles:
LGTM

Issue #230, opened 2017-05-16
Layers are shown twice when loading data from oq-engine outputs


ptormene:
The reference to the same layer is present in the Layers Panel both in the corresponding group and in the general group. It should be displayed only in the corresponding group.

ptormene:
The following commit caused the issue: https://github.com/gem/oq-irmt-qgis/commit/609c01622e2cc0e4c5c9fa6284f5e139dc1265bc
Unfortunately, if we revert this change we have a segmentation fault after all tests pass. I suppose it could be due to problems with layer groups.

ptormene:
It should be ok now

PR #231, opened 2017-05-17
Let the size of points depend on the zooming level

ptormene:
If I solve the issue https://github.com/gem/oq-irmt-qgis/issues/230 I cause a segmentation fault in tests, and I still haven't found a way to make it work. Also on Travis the segmentation fault occurs for both the ltr and the latest version of QGIS. The segfault disappears if we skip the test that loads multiple uhs layers. It is the only test that covers the functionality of the checkbox to load only the selected items or everything.


ptormene:
For now, I suppose it is reasonable to skip that single test and to fix #230. We will try to solve the segmentation fault in a future PR.

PR #237, opened 2017-05-18
Display the log of a OQ-Engine calculation in a dedicated dialog...

ptormene:
...instead of in the QGIS log window

Fixes https://github.com/gem/oq-irmt-qgis/issues/234


micheles:
If somebody closes the engine server (or it dies) you get this error:

```python
Traceback (most recent call last):
  File "/home/michele/.qgis2/python/plugins/svir/dialogs/show_console_dialog.py", line 55, in refresh_calc_log
    if calc_status['status'] in ('complete', 'failed'):
TypeError: 'NoneType' object has no attribute '__getitem__'
```
Perhaps we can do something better, like an error saying "Could not access the engine server" or such.

PR #277, opened 2017-06-12
Some integration tests

ptormene:
I changed things in order to avoid using any dialogs and creating any layers. This implied to re-write some functionalities that were already present in the dialogs (but that I simplified a lot here, for the simpler purposes).

Instead of loading the outputs as layers (which is tested in the unit tests), here I am just checking that the dtypes are as expected.

TODO: currently the name of the directory that contains the oq-engine demos is hardcoded, as well as the url and port of the connection with engine server. We will also have to set up Travis and/or Jenkins with a working engine server.

### DIFF
--- a/svir/test/integration/test_drive_oq_engine.py
+++ b/svir/test/integration/test_drive_oq_engine.py
@@ -40,10 +40,12 @@ class DriveOqEngineTestCase(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         cls.session = Session()
-        # FIXME: it might be passed as argument when running integration tests
+        # TODO: it might be passed as argument when running integration tests
         cls.demos_dir = os.path.join(
-            os.path.expanduser('~'), 'projects', 'oq-engine', 'demos')
-        cls.hostname = 'http://0.0.0.0:8800'
+            os.path.dirname(__file__),
+            os.pardir, os.pardir, os.pardir, os.pardir,
+            'oq-engine', 'demos')
+        cls.hostname = 'http://localhost:8800'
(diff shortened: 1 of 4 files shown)


svir/test/integration/test_drive_oq_engine.py
```diff
@@ -0,0 +1,234 @@
...
+
+class DriveOqEngineTestCase(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.session = Session()
+        # FIXME: it might be passed as argument when running integration tests
+        cls.demos_dir = os.path.join(
+            os.path.expanduser('~'), 'projects', 'oq-engine', 'demos')
+        cls.hostname = 'http://0.0.0.0:8800'
```
daniviga: `0.0.0.0` works only for a binding/listening address. Probably you want `localhost` here.

svir/test/integration/test_drive_oq_engine.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+class DriveOqEngineTestCase(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.session = Session()
+        # FIXME: it might be passed as argument when running integration tests
+        cls.demos_dir = os.path.join(
+            os.path.expanduser('~'), 'projects', 'oq-engine', 'demos')
```
daniviga: I would pass it as an argument and `oq-engine/demos` (relative to cwd) as default

ptormene:
Replaced by https://github.com/gem/oq-irmt-qgis/pull/282

PR #279, opened 2017-06-14
Load dmg_by_asset from npz

ptormene:
Closes https://github.com/gem/oq-irmt-qgis/issues/278

I replaced the loader from csv with this one from npz.

As in the case of losses_by_asset, the test do not check the contents of the created layers, but it checks that the dropdown menus are populated correctly and that the layers are built without raising exceptions.


ptormene:
Also closes https://github.com/gem/oq-irmt-qgis/issues/267

PR #282, opened 2017-06-15
Integration tests loading existing calculations

ptormene:
If you type `make integrationtest`, the tool iterates through all available calculations. For all outputs of each calculation, if a loader exists for that output type, the plugin attempts to load it. If any error occurs, the traceback is printed and the testing continues on all the remaining outputs. At the end of the loop, a short report listing the errors that occurred is printed (if present), otherwise tests complete successfully.

In the same PR, I am changing the presentation of the list of calculations, displaying the calculation mode instead of the job type. I am also using the calculation mode to hide the `Load as layer` button in case of gmf_data for an event based calculation.

Thanks to @daniviga, Travis loads a dump with oq-engine calculations already finished, starts the engine server and launches the integration tests.


svir/dialogs/drive_oq_engine_server_dialog.py
```diff
@@ -86,6 +86,16 @@
...
+OUTPUT_TYPE_LOADERS = {
+    'ruptures': LoadRupturesAsLayerDialog,
+    'dmg_by_asset': LoadDmgByAssetAsLayerDialog,
+    'gmf_data': LoadGmfDataAsLayerDialog,
+    'hmaps': LoadHazardMapsAsLayerDialog,
+    'hcurves': LoadHazardCurvesAsLayerDialog,
+    'uhs': LoadUhsAsLayerDialog,
+    'losses_by_asset': LoadLossesByAssetAsLayerDialog}
+assert set(OUTPUT_TYPE_LOADERS.keys()) == OQ_ALL_LOADABLE_TYPES
+
```
micheles: You do not need `.keys()` here

svir/utilities/utils.py
```diff
@@ -98,21 +99,26 @@ def log_msg(message, tag='GEM IRMT Plugin', level='I', message_bar=None,
...
-        elif level == 'C':
-            title = 'Error'
-            duration = duration if duration is not None else 0
-        message_bar.pushMessage(tr(title),
-                                tr(message),
-                                levels[level]['bar'],
-                                duration)
+
+    # if we are running nosetests, exit on critical errors
+    if 'nose' in sys.modules.keys() and level == 'C':
```
micheles: You do not need `.keys()` here

micheles:
LGTM

PR #284, opened 2017-06-20
Update links in .travis.yml following oq-engine/2908

daniviga:
Now using http://artifacts.openquake.org/travis/oqdata.zip

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_install:
     - sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-key 073D307A618E5811
     - sudo apt-get update -q
     - git clone --depth=1 -q https://github.com/gem/oq-engine.git
-    - curl -sO http://artifacts.openquake.org/demos/dump.zip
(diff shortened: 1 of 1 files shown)


daniviga:
```
FAIL: test_load_outputs (svir.test.integration.test_drive_oq_engine.LoadOqEngineOutputsTestCase)

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/home/travis/build/gem/oq-irmt-qgis/svir/test/integration/test_drive_oq_engine.py", line 159, in test_load_outputs

    % (len(calc_list), expected_num_calcs))

AssertionError: Found 15 calculations; expected 16
```

It seems that you'll need to update the tests @ptormene

micheles:
This happens because the Area Source demo is being run twice. After the fix in https://github.com/gem/oq-engine/pull/2904 we can probably avoid that: even if the demo looks failed now we can access its outputs.

ptormene:
On my machine, after running all demos, I have 16 calculations, including a single `Classical PSHA — Area Source`.

PR #285, opened 2017-06-20
Update documentation

ptormene:
After recent changes, the apidoc became very outdated and it was not possible to compile it anymore. I rebuilt it from scratch and I had to fix many docstrings that sphinx was complaining about.
I've added `make doc` to travis, so also the compilation of the manual and of the apidoc are tested.
I slightly modified the Makefile to make possible to build on travis only the html version of both the apidoc and the user manual (in order to avoid using latex on travis).


micheles:
LGTM

.travis.yml
```diff
@@ -34,6 +36,8 @@ before_script:
 script:
     - cd $TRAVIS_BUILD_DIR/svir && make test
     - cd $TRAVIS_BUILD_DIR/svir && make integrationtest
+    - cd $TRAVIS_BUILD_DIR/svir && make build_apidoc
```
daniviga: A minor comment. You may want to move the documentation generation in the `after_succes` block (see https://github.com/gem/oq-engine/blob/master/.travis.yml#L39)

PR #291, opened 2017-06-27
Load the modified hazard exports

ptormene:
Companion of https://github.com/gem/oq-engine/pull/2862

The big advantage of the new format is that now we don't need to store long json strings inside single cells of qgis layers. All cells will contain a scalar number. The drawback is that in layers for hcurves we will need to store in the field names the combination of IMT_IML, which can easily exceed the Shapefile limit of 10 characters. Therefore, layers will have to be saved in a different format (e.g., GeoJson).

The investigation time was made available engine-side, and it is used here postponing it to the name of each layer, adding it to the layer as a customProperty and showing it in the title of plots for hcurves and uhs.

Closes https://github.com/gem/oq-irmt-qgis/issues/288.


svir/dialogs/load_hcurves_as_layer_dialog.py
```diff
@@ -58,23 +57,32 @@ def set_ok_button(self):
         self.ok_button.setEnabled(
             bool(self.path) and self.rlz_cbx.currentIndex() != -1)
 
+    def populate_rlz_cbx(self):
+        self.rlzs = self.npz_file['all'].dtype.names[2:]  # ignore lon, lat
+        self.rlz_cbx.clear()
```
micheles: The name rlzs is misleading. I would use `rlz_or_stats`

svir/dialogs/load_hmaps_as_layer_dialog.py
```diff
@@ -59,11 +59,19 @@ def set_ok_button(self):
         self.ok_button.setEnabled(
             bool(self.path) and self.poe_cbx.currentIndex() != -1)
 
+    def populate_rlz_cbx(self):
+        # excluding lon, lat and the final vs30
+        self.rlzs = self.npz_file['all'].dtype.names[2:-1]
+        self.rlz_cbx.clear()
```
micheles: rlzs_or_stats

svir/dialogs/load_output_as_layer_dialog.py
```diff
@@ -443,6 +443,19 @@ def load_from_npz(self):
         # if self.npz_file is not None:
         #     self.npz_file.close()
 
+    def get_investigation_time(self):
+        try:
+            investigation_time = self.npz_file['investigation_time']
+        except KeyError:
+            investigation_time = None
```
micheles: Please add a comment that this happens for outputs without investigation_time. For hazard outputs is must exists and cannot be None.

svir/dialogs/load_output_as_layer_dialog.py
```diff
@@ -443,6 +443,19 @@ def load_from_npz(self):
...
+        else:
+            # check if it's integer
+            if numpy.equal(numpy.mod(investigation_time, 1), 0):
+                investigation_time = int(investigation_time)
+            else:
+                investigation_time = float(investigation_time)
+        return investigation_time
+
```
micheles: It is better not to hide the fact that the investigation time is a float. Let it to appear as a float, since it is a float.

svir/dialogs/viewer_dock.py
```diff
@@ -657,27 +651,46 @@ def on_export_data_button_clicked(self):
...
-            # write header
-            line = 'lon,lat,%s' % (
-                ','.join(map(str, list(self.current_abscissa))))
-            csv_file.write(line + os.linesep)
-
             if self.output_type == 'recovery_curves':
+                # write header
+                line = 'lon,lat,%s' % (
+                    ','.join(map(str, list(self.current_abscissa))))
+                csv_file.write(line + os.linesep)
```
micheles: Do we need the call to `list` here?

svir/dialogs/viewer_dock.py
```diff
@@ -657,27 +651,46 @@ def on_export_data_button_clicked(self):
...
+                    feature = next(self.active_layer.getFeatures(
+                        QgsFeatureRequest().setFilterFid(site)))
+                    lon = feature.geometry().asPoint().x()
+                    lat = feature.geometry().asPoint().y()
+                    line = '%s,%s,%s' % (lon, lat, poes)
+                    csv_file.write(line + os.linesep)
+            elif self.output_type == 'uhs':
+                # write header
+                line = 'lon,lat,%s' % (
+                    ','.join(map(str, list(self.current_abscissa))))
```
micheles: Idem as before

svir/utilities/utils.py
```diff
@@ -964,3 +964,22 @@ def import_layer_from_csv(parent,
...
 def listdir_fullpath(path):
     return [os.path.join(path, filename) for filename in os.listdir(path)]
+
+
+def get_param_from_comment_line(param_name, comment_line):
+    err_msg = 'Unable to extract param %s from line: %s' % (param_name,
+                                                            comment_line)
+    try:
+        # remove "# " from comment_line
+        comment_line = comment_line.split('# ')[1]
```
micheles: I would change this to

```
comment, rest = comment_line.split('# ')
```

micheles: Please add some doctests with failing cases

svir/utilities/utils.py
```diff
@@ -964,3 +964,47 @@ def import_layer_from_csv(parent,
...
+    params_dict = collections.OrderedDict()
+    param_defs = rest.split(', ')
+    for param_def in param_defs:
+        try:
+            name, value = param_def.split('=')
+        except ValueError:
+            raise LookupError(err_msg)
+        else:
+            params_dict[name] = value
+    return params_dict
```
micheles: The errors here are more ParsingError than LookupError.

micheles:
LGTM

PR #293, opened 2017-06-29
Install SAGA on Travis and check that SAGA version > 2.3

ptormene:
In recent versions of QGIS, versions of SAGA below 2.3 are not supported anymore (see https://gis.stackexchange.com/questions/245497/which-saga-versions-does-qgis-2-18-10-support).
Therefore, I added a check to use the fallback algorithm in case SAGA is found but its version is too old.
I am investigating from which version QGIS stopped supporting older versions of SAGA. The commit that introduced the change is https://github.com/qgis/QGIS/commit/732b501db16c14aaf569716d8f4418c437b39ec1


daniviga:
@ptormene SAGA for 16.04 and 14.04 is now available in our PPA https://launchpad.net/~openquake/+archive/ubuntu/saga

You can change `.travis.yml` and add the following lines before the main `apt install`

```
sudo add-apt-repository ppa:openquake/saga
sudo apt-get update
```

then you can add `saga python-saga` to the `apt install` command.

PR #298, opened 2017-07-17
Add descriptions and formulae of operators used to compute composite indices

ptormene:
The addition looks like this:
screenshot from 2017-07-17 14-31-49


pslh:
LGTM

PR #304, opened 2017-07-28
Plot statistical outputs together

ptormene:
The current behavior of the OQ-Engine is to produce statistical outputs (e.g. mean, quantiles, etc.) for calculations with multiple realizations. Through a multiselect widget, the user can now choose the statistics to plot together. We can therefore display curves/spectra corresponding to multiple sites and multiple statistics. By hovering on curves or on their legend items, the corresponding sites are highlighted in the map. The legend reports, for each curve, the geographical location (lon, lat) of the corresponding point and the realization (if single) or statistic.

WIP: Tests are green, but I want to stress the plugin a little more, to make sure I haven't broken anything (for instance, recovery modeling)


micheles:
Travis is unhappy!

micheles:
The plot window in the docker lists too many output types (i.e. the recovery_model or the UHS in calculations where there are no such outputs). However I would fix that in a separate pull request.

ptormene:
In the data viewer dock I could check if the active layer has a specific output type, preventing from selecting other incompatible output types. In case the active layer does not have an output type (it might happen for old layers or for layers that have not been directly built through our loaders) I could allow the user to select all the available output types, providing a clear error in case, after the selection, something goes wrong. I am going to add a ticket and to address it in another PR.

micheles:
LGTM

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/257

PR #306, opened 2017-08-16
Sum losses_by_asset and dmg_by_asset by zone

ptormene:
As requested by the risk team, I am simplifying the process of loading losses_by_asset and dmg_by_asset, giving the user a single widget to load the output and launch the aggregation of points by polygons. Before, it was possible to obtain the same result in two separate steps: loading the output, then opening a separate widget to select a loss layer and a zonal layer and run the aggregation.
I am having troubles with segfaults in tests, so I am skipping the 2 tests that I've added for this new feature. I still haven't found a solution for these (and other) segfaults.


ptormene:
Michele suggested the following changes:

1. add a combobox to select the field containing the id of each zone in the zonal layer with polygons
2. change the name of the field containing the number of points per zone

PR #310, opened 2017-09-06
Embed gem standalone web apps in the plugin

ptormene:
3 new buttons are added to the toolbar (new icons will be needed for them), opening corresponding widgets, each containing a web view with a OQ Standalone App embedded (IPT, TaxtWEB and Taxonomy). Whenever a page is requested, the plugin adds a header to the request, so the server-side application can be aware that it is being called by the plugin and it can behave differently. The plugin offers to the web applications a python api that can be called server-side to run actions inside QGIS. Therefore, the interaction between the plugin and the web applications can be bi-directional.
When a page attempts to open a url in a new window, the same window is used instead. We should further discuss and decide how the plugin should behave in this case.
Cookies are saved persistently in QSettings.


ptormene:
Draft of the workflow to launch OQ-Engine calculations from IPT:

1. IPT accumulates input files and exposes an API to download them
2. IPT calls the IRMT API to launch an OQ-Engine calculation
3. IRMT creates a folder, to be used to collect the input files
4. IRMT calls the IPT API to download the input files into the new folder
5. IRMT opens the dialog to select the input files for a OQ-Engine calculation, preselecting the input files downloaded from IPT
6. The user checks that all files are there, and presses the button to run the calculation

svir/dialogs/ipt_dialog.py
```diff
@@ -0,0 +1,291 @@
...
+        """
+        try:
+            ipt_dir = self.parent().ipt_dir
+            file_names = QFileDialog.getOpenFileNames(
+                self.parent().parent(), 'Select files', ipt_dir)
+            ls = [os.path.basename(file_name) for file_name in file_names]
+        except Exception as exc:
+            return {'ret': 1, 'reason': str(exc)}
+        else:
+            return {'ret': ls, 'reason': 'ok'}
```
nastasi-oq: Seems to me not consistent the return semantic, ret change type if success or not ... not so comfortable.
could be always an integer and 2 different other attribute in case of success and error ('content' and 'reason')

svir/dialogs/settings_dialog.py
```diff
@@ -240,4 +242,19 @@ def accept(self):
...
+            # refresh on all the registered items
+            if self.irmt_main.ipt_dlg is not None:
+                self.irmt_main.ipt_dlg.set_host()
+                self.irmt_main.ipt_dlg.load_homepage()
+            if self.irmt_main.taxtweb_dlg is not None:
+                self.irmt_main.taxtweb_dlg.set_host()
+                self.irmt_main.taxtweb_dlg.load_homepage()
+            if self.irmt_main.taxonomy_dlg is not None:
+                self.irmt_main.taxonomy_dlg.set_host()
+                self.irmt_main.taxonomy_dlg.load_homepage()
```
nastasi-oq: Add a loop here on `[self.irmt_main.ipt_dlg, self.irmt_main.taxtweb_dlg, self.irmt_main.taxonomy_dlg]`

svir/dialogs/standalone_app_dialog.py
```diff
@@ -0,0 +1,157 @@
...
+                 parent=None):
+        super(StandaloneAppDialog, self).__init__(parent)
+
+        self.message_bar = QgsMessageBar(self)
+        self.app_name = app_name
+        self.app_descr = app_descr
+        self.gem_header_name = gem_header_name
+        self.gem_header_value = gem_header_value
+        self.web_view = None
+        self.set_host()
```
nastasi-oq: Why `set_host()` and after each creation (search again set_host around) ?

svir/dialogs/ipt_dialog.py
ptormene: I agree

svir/dialogs/standalone_app_dialog.py
ptormene: I've added a comment in commit 1f7f740, to exaplain that the user might switch to a different engine server while having embedded applications already initialized and pointing to the previous server, therefore we must let all apps point to the new engine server.

svir/utilities/shared.py
```diff
@@ -240,3 +240,9 @@ def __repr__(self):
     experimental_enabled=False,
     developer_mode=False,
 )
+
+
```
nastasi-oq: Add a description to avoid to recollect 100x meaning after some week.

PR #311, opened 2017-09-08
Fix a bug occurring while redrawing curves in a corner case

ptormene:
The error occurred when performing this sequence of actions:
1) draw curves from a layer that uses the Data Viewer
2) select a layer that does not use the Data Viewer
3) draw curves from another layer (not the same used at number 1) that uses the Data Viewer.
I was caused by the fact that the object keeping track of the current selection was not reinitialized when switching the active layer to one that does not use the Data Viewer.

In addition to the fix, I've also made some minor changes that make the logic slightly more readable.

### DIFF
--- a/svir/dialogs/viewer_dock.py
+++ b/svir/dialogs/viewer_dock.py
@@ -623,10 +623,14 @@ class ViewerDock(QtGui.QDockWidget, FORM_CLASS):
         if (self.iface.activeLayer() is not None
                 and self.iface.activeLayer().type() == QgsMapLayer.VectorLayer
                 and self.iface.activeLayer().geometryType() == QGis.Point):
-            self.iface.activeLayer().selectionChanged.connect(self.redraw)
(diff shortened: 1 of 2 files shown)


svir/dialogs/viewer_dock.py
```diff
@@ -654,12 +658,15 @@ def layer_changed(self):
...
             if self.iface.activeLayer().selectedFeatureCount() > 0:
                 self.set_selection()
 
     def remove_connects(self):
         try:
-            self.iface.activeLayer().selectionChanged.disconnect(self.redraw)
+            self.iface.activeLayer().selectionChanged.disconnect(
+                self.set_selection)
         except (TypeError, AttributeError):
             pass
```
micheles: Could you add a comment saying when a TypeError is expected and when an AttributeError is expected?

micheles:
Apart the minor comment, LGTM

PR #312, opened 2017-09-18
Fix a key error exception in redrawing curves

ptormene:
A previous merge (https://github.com/gem/oq-irmt-qgis/pull/311) introduced a bug: a KeyError that occurred because of a missing reinitialization of the object holding the curves for the currently selected sites, in a corner case. Now the problem should be fixed. I've also renamed a method, to make it slightly more intuitive.


micheles:
Please add a test for the bug, if it takes a reasonable amount of effort.

PR #315, opened 2017-09-22
Strip /engine from hostname in user settings

ptormene:
If the the user copies/pastes the url from the webui into the plugin settings, the final `/engine` has to be removed from the url. I am stripping it before saving the user settings.


micheles:
LGTM

PR #317, opened 2017-09-26
Visualize agg_curves-rlzs and agg_curves-stats oq-engine outputs

ptormene:
For these outputs, I am leveraging the api introduced in gem/oq-engine#2929

Tests on Travis fail while attempting to load objects that were pickled by the oq-engine, which requires to import the oq-engine libraries.


ptormene:
I modified the loader, following the change proposed in https://github.com/gem/oq-engine/pull/3054/files

svir/dialogs/viewer_dock.py
```diff
@@ -284,6 +293,80 @@ def set_output_type_and_its_gui(self, new_output_type):
         # else.
         self.output_type = new_output_type
 
+    def load_agg_curves(self, calc_id, session, hostname, output_type):
+        self.change_output_type(output_type)
+        url = '%s/v1/calc/%s/extract/%s' % (hostname, calc_id, output_type)
+        resp_content = session.get(url).content
+        self.agg_curves = numpy.load(io.BytesIO(resp_content))
```
micheles: As more outputs will become available as npz, you will probably want an utility function `extract_npz(session, hostname, calc_id, output_type)`.

ptormene: I agree. I will make some refactoring in order to load different kinds of outputs. I will also probably re-arrange the classification and management of different kinds of output types.

svir/dialogs/viewer_dock.py
```diff
@@ -831,22 +929,49 @@ def write_export_file(self, filename):
                         line += "," + ",".join([
                             str(value) for value in values])
                     csv_file.write(line + os.linesep)
+            elif self.output_type == 'agg_curves-rlzs':
+                num_rlzs = self.agg_curves['array'].shape[1]
```
micheles: A comment on the shape of the array would be useful (it is (P, R) for rlzs and (P, S) for stats) being

P = number of return periods
R  = number of realizations
S = number of statistics

svir/dialogs/viewer_dock.py
```diff
@@ -831,22 +929,49 @@ def write_export_file(self, filename):
...
+                rlzs_str = ['rlz-%s' % rlz for rlz in range(num_rlzs)]
+                # write header
+                line = 'return_period,' + ','.join(rlzs_str)
+                csv_file.write(line + os.linesep)
+                for i, return_period in enumerate(
+                        self.agg_curves['return_periods']):
+                    values = self.agg_curves['array'][self.current_loss_type]
+                    line = str(return_period) + "," + ",".join(
+                        [str(value) for value in values[i]])
```
micheles: Don't you have a generic utility to write CSV files? Even using the csv module in the standard library could be an option.

ptormene: I agree it would be cleaner to use the csv module, although I am not so sure it would actually simplify things. But I guess it would at least improve readability.

svir/dialogs/viewer_dock.py
ptormene: Done

svir/dialogs/viewer_dock.py
ptormene: I've done some refactoring to have extract_npz as a general method to be used by different loaders.

svir/dialogs/viewer_dock.py
ptormene: Done

svir/test/data/hazard/hazard_curves_SA(0.1).csv
```diff
@@ -1,3 +1,3 @@
...0.1)_0.397,quantile-0.05_SA(0.1)_0.556,quantile-0.05_SA(0.1)_0.778,quantile-0.05_SA(0.1)_1.09,quantile-0.05_SA(0.1)_1.52,quantile-0.05_SA(0.1)_2.13,quantile-0.5_SA(0.1)_0.005,quantile-0.5_SA(0.1)_0.007,quantile-0.5_SA(0.1)_0.0098,quantile-0.5_SA(0.1)_0.0137,quantile-0.5_SA(0.1)_0.0192,quantile-0.5_SA(0.1)_0.0269,quantile-0.5_SA(0.1)_0.0376,quantile-0.5_SA(0.1)_0.0527,quantile-0.5_SA(0.1)_0.0738,quantile-0.5_SA(0.1)_0.103,quantile-0.5_SA(0.1)_0.145,quantile-0.5_SA(0.1)_0.203,quantile-0.5_SA(0.1)_0.284,quantile-0.5_SA(0.1)_0.397,quantile-0.5_SA(0.1)_0.556,quantile-0.5_SA(0.1)_0.778,quantile-0.5_SA(0.1)_1.09,quantile-0.5_SA(0.1)_1.52,quantile-0.5_SA(0.1)_2.13,quantile-0.95_SA(0.1)_0.005,quantile-0.95_SA(0.1)_0.007,quantile-0.95_SA(0.1)_0.0098,quantile-0.95_SA(0.1)_0.0137,quantile-0.95_SA(0.1)_0.0192,quantile-0.95_SA(0.1)_0.0269,quantile-0.95_SA(0.1)_0.0376,quantile-0.95_SA(0.1)_0.0527,quantile-0.95_SA(0.1)_0.0738,quantile-0.95_SA(0.1)_0.103,quantile-0.95_SA(0.1)_0.145,quantile-0.95_SA(0.1)_0.203,quantile-0.95_SA(0.1)_0.284,quantile-0.95_SA(0.1)_0.397,quantile-0.95_SA(0.1)_0.556,quantile-0.95_SA(0.1)_0.778,quantile-0.95_SA(0.1)_1.09,quantile-0.95_SA(0.1)_1.52,quantile-0.95_SA(0.1)_2.13
```
micheles: The problem with the CSV tests is they are very fragile. On different platforms the numbers could be slightly different. With different versions of shapely/libgeos/numpy/etc the numbers could be slightly different. The parameters of the demos could change in the future and the numbers (even the number of rows and columns) could change. I suggest to simply check that a file is generated with at least two rows and three columns.

ptormene: Done

svir/dialogs/viewer_dock.py
```diff
@@ -776,34 +879,31 @@ def on_export_data_button_clicked(self):
                     '~/recovery_curves_%s.csv' %
                     self.approach_cbx.currentText()),
                 '*.csv')
-        if filename:
+        if filename is not None:
             self.write_export_file(filename)
 
     def write_export_file(self, filename):
```
micheles: In the engine when we export CSV files, we use an extension `_XYZ.csv` where XYZ is the calculation ID. This could be a suggestion for a future PR.

micheles:
LGTM

svir/test/unit/test_load_oq_engine_output_as_layer.py
```diff
@@ -271,20 +271,31 @@ def test_load_dmg_by_asset_aggregate_by_zone(self):
...
         layer.select([1, 2])
         # probably we have the wrong layer selected (uhs produce many layers)
         self.viewer_dock.write_export_file(exported_file_path)
-        expected_file_path = os.path.join(
-            self.data_dir_name, 'hazard', expected_file_name)
-        self.assertTrue(
-            filecmp.cmp(exported_file_path, expected_file_path),
-            'The exported file (%s) is different with respect to the'
-            ' reference one (%s)' % (exported_file_path, expected_file_path))
+        with open(exported_file_path, 'r') as got:
```
micheles: You could add a comment about why we are doing such relaxed check, for future memory.

svir/dialogs/viewer_dock.py
ptormene: I've just opened an issue to keep track of this suggestion

svir/test/unit/test_load_oq_engine_output_as_layer.py
ptormene: Done

PR #321, opened 2017-10-02
Load the dmg_total OQ-Engine output

ptormene:
When loading the `dmg_total` output of a scenario damage calculation, the Data Viewer lets the user select one of the realizations and one of the loss types, and there is also a checkbox to exclude (or keep) the visualization of "no damage" (the number of assets that has no damage is usually much greater than the number of assets in each of the other damage states). A bar plot is shown, displaying the number of assets in each damage state, for the selected realization and loss type.

I have not implemented the export functionality for this output, because it does not look so useful, since it is possible to download a quite simple csv directly from the OQ-Engine. If required, we can implement this in another PR. Anirudh agreed it is not needed now.

NOTE: OQ-Engine side, the outputs `dmg_by_asset`, `dmg_by_tag` and `dmg_total` are probably going to be unified and called just `dmg`. As a consequence, this will imply to change the workflow that we have already implemented for `dmg_by_asset` and `dmg_total`.

### DIFF
--- a/svir/dialogs/viewer_dock.py
+++ b/svir/dialogs/viewer_dock.py
@@ -345,14 +345,15 @@ class ViewerDock(QDockWidget, FORM_CLASS):
         rlzs = ['rlz-%s' % rlz for rlz in range(num_rlzs)]
         self.rlz_cbx.blockSignals(True)
         self.rlz_cbx.clear()
-        self.rlz_cbx.blockSignals(False)
         self.rlz_cbx.addItems(rlzs)
+        self.rlz_cbx.blockSignals(False)
 
         loss_types = self.dmg_total['array'].dtype.names
         self.loss_type_cbx.blockSignals(True)
         self.loss_type_cbx.clear()
-        self.loss_type_cbx.blockSignals(False)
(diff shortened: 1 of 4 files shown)


micheles:
LGTM!

svir/dialogs/viewer_dock.py
```diff
@@ -374,6 +430,43 @@ def draw_agg_curves(self, output_type):
...
+        else:
+            color = 'IndianRed'
+        self.plot.clear()
+        self.plot.bar(indX+padding_left, height=means, width=bar_width,
+                      yerr=stddevs, error_kw=error_config, color=color,
+                      linewidth=1.5, alpha=0.6)
+        self.plot.set_title('Damage distribution')
+        self.plot.set_xlabel('Damage state')
+        self.plot.set_ylabel('Number of assets in damage state')
+        # self.plot.set_xticks(indX+padding_left+bar_width/2., dmg_states)
```
micheles: Why this is commented?

ptormene: I forgot to remove it. Done now.

PR #323, opened 2017-10-05
Visualize scenario damage by tag

ptormene:
The user can select a realization, a loss type and many tags (only one value per tag can be selected), and a bar chart is plotted, displaying the number of assets in each damage state. Also the "Export data" functionality is implemented.

The ListMultiSelectMonoWidget is a subclass of ListMultiSelectWidget, that enables to select only one tag value at a time. If you select another value, it is swapped with the one that was previously selected. I could have used combo boxes instead, but the current solution will make it very easy, in the future, to enable the tool to select multiple values per tag (if the engine will allow to filter by multiple values per tag).

Error bars are not displayed, because the engine is currently unable to provide standard deviations for the selected data. I've added some comments, that will be useful if the engine will provide this functionality.

### DIFF
--- a/svir/dialogs/viewer_dock.py
+++ b/svir/dialogs/viewer_dock.py
@@ -66,6 +66,7 @@ from svir.utilities.utils import (get_ui_class,
 from svir.recovery_modeling.recovery_modeling import (
     RecoveryModeling, fill_fields_multiselect)
 from svir.ui.list_multiselect_widget import ListMultiSelectWidget
+from svir.ui.list_multiselect_mono_widget import ListMultiSelectMonoWidget
 
 from svir import IS_SCIPY_INSTALLED
 
@@ -281,6 +282,8 @@ class ViewerDock(QDockWidget, FORM_CLASS):
             QAbstractItemView.SingleSelection)
         self.tag_names_multiselect.unselected_widget.setSelectionMode(
             QAbstractItemView.SingleSelection)
+        self.tag_names_multiselect.select_all_btn.hide()
+        self.tag_names_multiselect.deselect_all_btn.hide()
         self.typeDepVLayout.addWidget(self.tag_names_multiselect)
         self.tag_names_multiselect.unselected_widget.itemClicked.connect(
             self.populate_tag_values_multiselect)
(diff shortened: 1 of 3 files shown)


svir/dialogs/viewer_dock.py
```diff
@@ -262,14 +265,113 @@ def create_fields_multiselect(self):
...
+        self.typeDepVLayout.addWidget(self.tag_values_multiselect)
+        self.tag_values_multiselect.selection_changed.connect(
+            self.update_selected_tag_values)
+
+    def populate_tag_values_multiselect(self, item):
+        tag_name = item.text()
+        self.current_tag_name = tag_name
+        self.tag_values_multiselect.setTitle(
+            'Select values for tag: %s' % tag_name)
+        selected_tag_values = [
```
micheles: You may want to give a name to `self.tags[tag_name]['values']`

svir/dialogs/viewer_dock.py
```diff
@@ -434,15 +561,28 @@ def draw_dmg_total(self):
...
+            self.plot.clear()
+            self.plot_canvas.draw()
+            return
+        rlz = self.rlz_cbx.currentIndex()
+        # TODO: re-add error bars when stddev will become available again
+        # loss_type = self.loss_type_cbx.currentText()
+        # means = self.dmg_total['array'][rlz][loss_type]['mean']
+        # stddevs = self.dmg_total['array'][rlz][loss_type]['stddev']
+        means = self.dmg_total['array'][rlz]
+        if any(means < 0):
```
micheles: Just for your information: since `means` is a numpy array, it is more efficient to use `(means < 0).any()` (not that it will make any difference in this case).

svir/dialogs/viewer_dock.py
```diff
@@ -453,8 +593,11 @@ def draw_dmg_total(self):
         else:
             color = 'IndianRed'
         self.plot.clear()
+        # self.plot.bar(indX, height=means, width=bar_width,
+        #               yerr=stddevs, error_kw=error_config, color=color,
+        #               linewidth=1.5, alpha=0.6)
```
micheles: Why this is commented out?

svir/dialogs/viewer_dock.py
ptormene: Done

svir/dialogs/viewer_dock.py
ptormene: Done

svir/dialogs/viewer_dock.py
ptormene: I've added a comment

PR #324, opened 2017-10-06
Use 2 buttons to load the same output type in different ways

ptormene:
`dmg_total` is going to be excluded from the available OQ-Engine outputs, and the corresponding functionalities will be conceptually included in the `dmg_by_asset` output. Therefore, `dmg_by_asset` needs to have 2 different buttons: the first one loads the output as layer; the second one displays bar charts interrogating the OQ-Engine through the new `extract` API. Before this PR, the plugin assumed that each engine output could be read in only one way. I changed the logic to obtain, from one single output type, also a modified version (e.g. from `dmg_by_asset`, also a `dmg_by_asset_show`) that can be handled separately.

### DIFF
--- a/svir/dialogs/drive_oq_engine_server_dialog.py
+++ b/svir/dialogs/drive_oq_engine_server_dialog.py
@@ -614,7 +614,7 @@ class DriveOqEngineServerDialog(QDialog, FORM_CLASS):
                 if not (row['type'] == 'gmf_data'
                         and 'event_based' in calculation_mode):
                     num_actions += 1  # needs additional column for loader btn
-            if "%s_show" % row['type'] in OQ_NO_MAP_TYPES:
+            if "%s_aggr" % row['type'] in OQ_NO_MAP_TYPES:
                 num_actions += 1
             max_actions = max(max_actions, num_actions)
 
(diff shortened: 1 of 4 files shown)


svir/dialogs/drive_oq_engine_server_dialog.py
```diff
@@ -631,23 +634,29 @@ def show_output_list(self, output_list, calculation_mode):
...
-                else:
-                    action = 'Load as layer'
-                # TODO: remove check when gmf_data will be loadable also for
-                #       event_based
-                if (output['type'] == 'gmf_data'
-                        and calculation_mode == 'event_based'):
-                    continue
-                button = QPushButton()
-                self.connect_button_to_action(
-                    button, action, output, outtype)
```
ptormene: I changed the indentation here, although it worked also the other way. I think it's logically clearer now.

micheles:
Now what was called `output_type` should probably be called `visualization_type`. Not sure if we should change the name for the moment, though.

micheles:
The name of the button should not be `Show`. I would like a name like `Aggregate`

micheles:
Apart from the naming suggestions, LGTM

PR #325, opened 2017-10-11
Visualize aggregated losses_by_asset

ptormene:
A second button ("Aggregate") is added for OQ-Engine outputs of type losses_by_asset.
In the Data Viewer, it is possible to select the loss type and a set of tags.
Only for one tag name, it is possible to select a `*` tag value, meaning that losses have to be aggregated for each tag value separately. It is also possible to select a single tag value for multiple tag names.
The result is a table in which each column corresponds to one of the available gsims, and there are 3 possible cases:

1. if only one tag name is selected, and value `*` is chosen, each row corresponds to one of the tag values for that tag name

2. if many tag names are selected, with a single value for each, the table will contain a single row, showing the aggregated losses for that combination

3. If many tag names are chosen, and for one of them the `*` value is chosen, each row of the table will display the aggregated losses for one of the values for the starred tag, compliant with the constrains given by the single values for the other tags

If the constraints produce an empty interception, the corresponding table rows will be empty.

### DIFF
--- a/svir/dialogs/viewer_dock.py
+++ b/svir/dialogs/viewer_dock.py
@@ -49,6 +49,8 @@ from qgis.PyQt.QtGui import (QColor,
                              QDockWidget,
                              QFileDialog,
                              QAbstractItemView,
+                             QTableWidget,
+                             QTableWidgetItem,
                              )
 from qgis.gui import QgsVertexMarker
 from qgis.core import QGis, QgsMapLayer, QgsFeatureRequest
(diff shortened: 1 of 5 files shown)


micheles:
In simple selection (i.e. no `*`) there is no need to visualize the tag column. Also there is no need to export it in CSV, There are also some pickle issues with Python 2 <-> 3, but they are not on your side, I need to investigate why and how pickle is used when creating the npz files.

PR #327, opened 2017-10-16
Use the same layer name for scenario_risk and scenario_damage

ptormene:
As requested by Cata, layers for gmfs should not be named `scenario_damage_gmfs_rlz-RLZ_eid-EID`, but, more generically, `scenario_gmfs_rlz-RLZ_eid-EID` (which is good also for `scenario_risk`)


ptormene:
See https://github.com/gem/oq-irmt-qgis/issues/326

micheles:
LGTM

PR #330, opened 2017-10-19
Fix curl and artifacts url in travis yaml

daniviga:
Moved to https and added `-L` to allow changes to the url in the future


daniviga:
Tests are green, merging

PR #334, opened 2017-11-16
Add multiselect for rlzs or stats in visualization of aggregated loss curves

ptormene:
This was requested by Cata, to have consistency with the visualization of hazard curves


micheles:
The tests are broken. It is due to a recent change in the engine https://github.com/gem/oq-engine/pull/3239. Here is the fix: https://github.com/gem/oq-engine/pull/3255.

PR #341, opened 2017-11-21
Use default credentials when settings are missing / Better display error messages

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/339

Please @daniviga check if it also fixes https://github.com/gem/oq-irmt-qgis/issues/338


svir/utilities/utils.py
```diff
@@ -1031,3 +1031,23 @@ def warn_scipy_missing(message_bar):
     msg = ("This functionality requires scipy. Please install it"
            " and restart QGIS to enable it.")
     log_msg(msg, level='C', message_bar=message_bar)
+
+
+def get_credentials(server):
+    """
+    Get from the QSettings the credentials to access the OpenQuake Platform
```
daniviga: I guess it's not only for the Platform

ptormene: Yes, I forgot to change the docstring, thanks

PR #342, opened 2017-11-21
If error occurs on importing matplotlib, warn about possible 64bit QGIS issues

ptormene:
Add a warning about the following issue: https://github.com/gem/oq-irmt-qgis/issues/224

### DIFF
--- a/svir/dialogs/viewer_dock.py
+++ b/svir/dialogs/viewer_dock.py
@@ -38,9 +38,9 @@ try:
     from matplotlib.lines import Line2D
 except ImportError as exc:
     raise ImportError(
-        'There was a problem importing matplotlib. It could happen when using'
-        ' a 64 bit version of QGIS. We recommend to use'
-        ' a 32 bit version instead: %s' % exc)
+        'There was a problem importing matplotlib. If you are using'
+        ' a 64bit version of QGIS on Windows, please try using'
+        ' a 32bit version instead. %s' % exc)
 
 from qgis.PyQt.QtCore import pyqtSlot, QSettings, Qt
 from qgis.PyQt.QtGui import (QColor,
--- a/svir/dialogs/viewer_dock.py
+++ b/svir/dialogs/viewer_dock.py
@@ -29,13 +29,18 @@ import numpy
 import io
 from collections import OrderedDict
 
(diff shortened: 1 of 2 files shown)


svir/dialogs/viewer_dock.py
```diff
@@ -29,13 +29,18 @@
...
+try:
+    from matplotlib.figure import Figure
+    from matplotlib.backends.backend_qt4agg import (
+        FigureCanvasQTAgg as FigureCanvas,
+        NavigationToolbar2QT as NavigationToolbar
+    )
+    from matplotlib.lines import Line2D
+except ImportError as exc:
+    raise ImportError(
+        'There was a problem importing matplotlib. It could happen when using'
```
pslh: Minor EN suggestion:
"There was a problem importing matplotlib, if you are using a 64bit version of QGIS with on Windows, please try using a 32bit version instead."

svir/recovery_modeling/recovery_modeling.py
```diff
@@ -26,7 +26,13 @@
 import os
 import json
 import bisect
-import matplotlib.pyplot as plt
+try:
+    import matplotlib.pyplot as plt
+except ImportError as exc:
+    raise ImportError(
+        'There was a problem importing matplotlib. It could happen when using'
```
pslh: See message suggestion above

pslh:
Testing on Windows 10 with QGIS 64bit 2.18.14 we see that, surprisingly, the matplotlib import does not fail, so the warning message is not displayed.  On the contrary it is possible to display hazard curves and recovery curves correctly.

daniviga:
@pslh have you tried to close QGIS and reopen it with the plug-in installed?
As I was trying to explain at scrum for some users it worked the first time they installed it. But then it failed with the Tkinter error after restarting QGIS; uninstalling  an reinstalling it did not fixed the issue.
That's way I was sceptical about an easy way to detect the issue at installation time, unless we check if we are running on Windows + 64bit

pslh:
@daniviga  yes, repeatedly.  We also removed the plugin and installed from git.  Using the latest 64bit version of QGIS today I have not been able to reproduce the bug at all, although I am sure I have seen the problem before (and so returned to 32bit).

PR #343, opened 2017-11-21
Fix check of layer capabilities, to work regardless from the locale

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/340


mbernasocchi:
lgtm. in effetti si usa sempre capabilities. non avevo mai visto capabilitiesstring

PR #349, opened 2017-12-27
Add partial description of risk outputs visualization

ptormene:
I've started documenting the visualization of OQ-Engine risk outputs, but it will require support from the scientific team, especially for what concerns the description of those outputs.


ptormene:
I propose to start merging this, and then add more in a future PR.

PR #367, opened 2018-01-31
Import Qt module from qgis.PyQt.QtCore instead of from qgis.PyQt

ptormene:
A Qt module exists also under qgis.PyQt and it is similar to the one contained in qgis.PyQt.QtCore, but:
1) in some cases it works on Ubuntu and not on macOS
2) it does not have exactly the same properties that are in the one under QtCore
Using always the Qt module contained in QtCore, I've tested it on Ubuntu and on macOS, and it works. Therefore, I'm removing the ugly workaround introduced in https://github.com/gem/oq-irmt-qgis/pull/363

### DIFF
--- a/svir/ui/list_multiselect_widget.py
+++ b/svir/ui/list_multiselect_widget.py
@@ -13,9 +13,8 @@ __author__ = 'marco@opengis.ch'
 __revision__ = '$Format:%H$'
 __date__ = '9/07/2013'
 
-# FIXME: Use wrapper as soon as it works also on Mac OS
-from PyQt4 import QtGui, QtCore, Qt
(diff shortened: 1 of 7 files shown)


svir/ui/list_multiselect_widget.py
```diff
@@ -133,7 +131,7 @@ def _setupUI(self):
         self._set_list_widget_defaults(self.unselected_widget)
         unselected_label = QtGui.QLabel()
         unselected_label.setText('Unselected')
-        unselected_label.setAlignment(Qt.Qt.AlignCenter)
+        unselected_label.setAlignment(QtCore.Qt.AlignCenter)
```
mbernasocchi: ah, I hadn't noticed this one...

PR #379, opened 2018-02-12
Remove menu items to load hmaps hcurves and uhs from npz

ptormene:
Those outputs can not be loaded from npz files anymore, but I forgot to remove the corresponding menu items (that became broken after that change). Now the only way to load those outputs is through the dialog that drives the OQ-Engine.


micheles:
LGTM, this is something I noticed and forgot to tell you about!

PR #384, opened 2018-02-14
Show loading progress

ptormene:
While attempting to address https://github.com/gem/oq-irmt-qgis/issues/383, I improved the feedback provided by the widget embedding OQ web applications, and I added a button to force reloading the homepage if needed (a confirmation dialog prevents from accidental clicks that would discard all the changes made in the page by the user).
Leveraging signals emitted by the page, I added a bar that displays the progress while loading a page. In case a page can not be loaded, an error message is shown.

Open issue:
still, if the engine server is down and you attempt to load a page, and afterwards you restart the engine server and do right-click -> reload, nothing happens. I still have not been able to fix it. Anyway, having the possibility to press a button to reload the homepage seems to be an acceptable workaround.


svir/dialogs/standalone_app_dialog.py
```diff
@@ -31,6 +31,9 @@
                               )
 from qgis.PyQt.QtGui import (QDialog,
                              QVBoxLayout,
+                             QPushButton,
+                             QHBoxLayout,
+                             QMessageBox,
```
nastasi-oq: Maybe a newline just before `QDialog, ` allow a more compact layout later ?

svir/ui/gem_qwebview.py
```diff
@@ -35,10 +35,14 @@
...
                               pyqtSlot,
+                              QUrl,
                               )
 from qgis.PyQt.QtGui import QSizePolicy, QFileDialog
 from svir.utilities.shared import DEBUG, REQUEST_ATTRS
-from svir.utilities.utils import tr
+from svir.utilities.utils import (tr,
+                                  create_progress_message_bar,
+                                  clear_progress_message_bar,
+                                  )
```
nastasi-oq: newline before `tr` and a more compact layout ?

ptormene: It's like this almost everywhere in the plugin. I prefer to keep it as it is, instead of changing it in the whole repository.

PR #393, opened 2018-02-23
Add calc_id to layers as customProperty and put it in exported filename

ptormene:
1. also layers that have nothing to do with the Data Viewer will contain the calc_id as a custom property, that can be retrieved through the Python Console (calc_id = iface.activeLayer().customProperty('calc_id')). Perhaps I should also provide a more intuitive way for the user to get the calc_id (for instance, a contextual menu, or I could add it to the layer name before the investigation time)
2. while exporting curves with the Data Viewer, the exported file name will contain the calc_id (fixes https://github.com/gem/oq-irmt-qgis/issues/320)


micheles:
I tried it on the China computation on cluster2 (number 17586) and it works but then when I opened the save CSV files it contained all sites not only the selected one.

PR #399, opened 2018-03-09
Fix reading damage states

ptormene:
Needed after the changes introduced in https://github.com/gem/oq-engine/pull/3505


daniviga:
I would open a ticket about the skipped tests, this could be a problem in the real world too (for example on Windows or macOS, trying to get the same of outputs)

ptormene:
https://github.com/gem/oq-irmt-qgis/issues/402

PR #411, opened 2018-03-27
Improve error message in case the engine is not running

ptormene:
Partially addressing https://github.com/gem/oq-irmt-qgis/issues/409


svir/dialogs/drive_oq_engine_server_dialog.py
```diff
@@ -846,7 +846,7 @@ def _handle_exception(self, exc):
                 err_msg += ' (you could try prepending http:// or https://)'
             elif isinstance(exc, SvNetworkError):
                 err_msg += (
-                    ' (please make sure the OpenQuake Engine Server is'
+                    ' (please make sure the OpenQuake Engine Server (WebUI) is'
                     ' running, and the username and password are'
                     ' spelled correctly)')
```
micheles: Just write `WebUI`

PR #418, opened 2018-03-28
Load realizations and sourcegroups from csv as layer

ptormene:
Addresses https://github.com/gem/oq-irmt-qgis/issues/413


micheles:
I do not like this approach Paolo. Having one different class for each csv output is a no go. There should be a single class and when you instantiate it you can pass to the instance the output_type, i.e. "sourcegroups" or "realizations". Otherwise you will have a lot of duplication.

ptormene:
I started with this approach, because the loader for ruptures is different, and I was expecting the other one for aggregate loss tables would have been different too.
I can unify these 2 equal loaders into a generic one, that can be reused also for other output types that do not require any additional processing.

svir/dialogs/load_realizations_as_layer_dialog.py
```diff
@@ -0,0 +1,65 @@
+# -*- coding: utf-8 -*-
+# /***************************************************************************
+# Irmt
+#                                 A QGIS plugin
+# OpenQuake Integrated Risk Modelling Toolkit
+#                              -------------------
+#        begin                : 2013-10-24
+#        copyright            : (C) 2014 by GEM Foundation
```
daniviga: 2018

svir/dialogs/load_sourcegroups_as_layer_dialog.py
```diff
@@ -0,0 +1,65 @@
+# -*- coding: utf-8 -*-
+# /***************************************************************************
+# Irmt
+#                                 A QGIS plugin
+# OpenQuake Integrated Risk Modelling Toolkit
+#                              -------------------
+#        begin                : 2013-10-24
+#        copyright            : (C) 2014 by GEM Foundation
```
daniviga: 2018

ptormene:
I've unified the loaders for simple csv outputs with no geometries, and I've added an utility to display the size of the csv file from which we are loading those simple outputs or more complex ones such as ruptures.

micheles:
A comment: by clicking on Load Layer I would expect the attribute table to pop up.

ptormene:
Done

PR #419, opened 2018-03-29
Set the style of layers imported from the OQ-Engine

ptormene:
Addresses a point of https://github.com/gem/oq-irmt-qgis/issues/326

For layers imported from the OQ-Engine, instead of using a gradient color ramp with the colors specified in the user settings, we will use an inverted spectral color ramp.
The number of classes will be the one specified in the user settings, plus one for values that are exactly zero.
The mode will be "natural breaks" for damage_by_asset and losses_by_asset, and "equal intervals" for ground motion fields and hazard maps (or the mode specified in the user settings, otherwise)


CatalinaYepes:
The suggestion for the default colors is:

- The color scheme for ground motion fields and hazard maps can be: 
_color map "spectral" inverted, 7 classes with "Equal interval"_

- The color scheme for "losses_by_asset" and "damage_by_asset" and other risk maps can be: 
color map "Reds", 7 classes with "Natural breaks"

ptormene:
screenshot from 2018-03-29 11-18-18
screenshot from 2018-03-29 11-17-22

svir/dialogs/load_output_as_layer_dialog.py
```diff
@@ -404,18 +404,34 @@ def style_maps(self):
...
+        # in some cases, we override the user-specified setting, and use
+        # instead a setting that was required by scientists
+        if self.output_type in ('dmg_by_asset', 'losses_by_asset'):
+            mode = QgsGraduatedSymbolRendererV2.Jenks  # jenks = natural breaks
+            ramp_type_idx = default_color_ramp_names.index('Reds')
+            ramp = default_qgs_style.colorRamp(
+                default_color_ramp_names[ramp_type_idx])
+            inverted = False
+        elif self.output_type in ('hmaps', 'gmf_data'):
+            mode = QgsGraduatedSymbolRendererV2.EqualInterval
```
daniviga: LGTM but let's use the original default with user's colors for cases not evaluated by ifs

ptormene:
image

ptormene:
I'm using "equal intervals", because pretty breaks (for some reason) produces an arbitrary number of classes (for instance, you set 7 classes, and you obtain 12 classes, including the one for zeros)

PR #422, opened 2018-03-30
Do not include AVG and NUM_POINTS when aggregating by zone

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/415

With an additional parameter, it is possible to specify if all the statistics have to be added to the zonal layer, or only sums. While aggregating damage and loss outputs from the engine, the new `sum_only` parameter is set to True.


svir/calculations/aggregate_loss_by_zone.py
```diff
@@ -70,7 +70,8 @@ def calculate_zonal_stats(loss_layer,
                           zone_id_in_zones_attr_name,
                           iface,
                           force_saga=False,
-                          force_fallback=False):
+                          force_fallback=False,
+                          sum_only=False):
     """
```
micheles: Instead of a flag `sum_only=False` I would use a flag `extra=True`. It is more readable to see something like `if extra:..`
instead of `if not sum_only: ...` Notice that you always use the negative form.

ptormene: Done

PR #425, opened 2018-04-03
Add loaders for Average Asset Losses Statistics

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/414

Now you can load the output as layer, and you can visualize the aggregated values in the Data Viewer


svir/dialogs/viewer_dock.py
```diff
@@ -607,6 +640,37 @@ def load_losses_by_asset_aggr(
...
+        self.loss_type_cbx.blockSignals(True)
+        self.loss_type_cbx.clear()
+        self.loss_type_cbx.addItems(loss_types)
+        self.loss_type_cbx.blockSignals(False)
+
+        to_extract = 'agglosses/%s' % loss_types[0]
+        npz = extract_npz(session, hostname, calc_id, to_extract,
+                          message_bar=self.iface.messageBar())
+        if npz['stats'] == numpy.array(None):
+            self.stats = ['rlz-0']
```
micheles: Perhaps a comment saying that this happens when there is a single realization is in order.

svir/dialogs/viewer_dock.py
```diff
@@ -785,6 +849,31 @@ def draw_losses_by_asset_aggr(self):
...
+    def draw_avg_losses_stats_aggr(self):
+        self.plot_canvas.hide()
+        clear_widgets_from_layout(self.table_layout)
+        losses_array = self.avg_losses_stats_aggr['array']
+        losses_array = self._to_2d(losses_array)
+        tags = None
+        try:
+            tags = self.avg_losses_stats_aggr['tags']
+        except KeyError:
+            pass
```
micheles: A comment says when the `pass` case happens would be useful. Why the 'tags' are missing sometimes?

ptormene: I found out that tags are needed (and available) only when the '*' is selected. In that case, we display them in the headers of the table, as row names. I've added some comments to clarify this.

PR #436, opened 2018-04-10
Style ruptures by trt or by magnitude

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/435


CatalinaYepes:
To style by magnitude, it will be better to classify the scale with `Pretty Breaks`, and remove the interval for zeros.

PR #441, opened 2018-04-17
Loading GMFs, show GMPEs instead of realizations

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/410

### DIFF
--- a/svir/dialogs/load_gmf_data_as_layer_dialog.py
+++ b/svir/dialogs/load_gmf_data_as_layer_dialog.py
@@ -68,12 +68,16 @@ class LoadGmfDataAsLayerDialog(LoadOutputAsLayerDialog):
         self.populate_gmpe_cbx()
         self.show_num_sites()
 
+    def _format_rlz(self, gmpe_idx):
+        # NOTE: assuming that rlzs are always in the format rlz-XXX
+        return "rlz-%03d" % gmpe_idx
+
     def show_num_sites(self):
         # NOTE: we are assuming all realizations have the same number of sites,
         #       which currently is always true.
         #       If different realizations have a different number of sites, we
         #       need to move this block of code inside on_rlz_or_stat_changed()
-        rlz = "rlz-%03d" % self.gmpe_cbx.currentIndex()
+        rlz = self._format_rlz(self.gmpe_cbx.currentIndex())
         gmpe_data = self.npz_file[rlz]
         self.num_sites_lbl.setText(
             self.num_sites_msg % gmpe_data.shape)
@@ -86,11 +90,12 @@ class LoadGmfDataAsLayerDialog(LoadOutputAsLayerDialog):
         self.gmpe_cbx.clear()
         self.gmpe_cbx.setEnabled(True)
         self.gmpe_cbx.addItems(self.gmpes)
-        self.rlzs_or_stats = ["rlz-%03d" % i for i in range(len(self.gmpes))]
-        self.gmpe_to_rlz = dict(zip(self.gmpes, self.rlzs_or_stats))
+        self.rlzs_or_stats = [
+            self._format_rlz(gmpe_idx) for gmpe_idx in range(len(self.gmpes))]
+        self.gmpe2rlz = dict(zip(self.gmpes, self.rlzs_or_stats))
 
     def on_gmpe_changed(self):
-        rlz = "rlz-%03d" % self.gmpe_cbx.currentIndex()
+        rlz = self._format_rlz(self.gmpe_cbx.currentIndex())
         self.dataset = self.npz_file[rlz]
         imts = self.dataset.dtype.names[2:]
         self.imt_cbx.clear()
@@ -118,13 +123,13 @@ class LoadGmfDataAsLayerDialog(LoadOutputAsLayerDialog):
                 continue
             with WaitCursorManager('Creating layer for "%s"...'
                                    % gmpe, self.iface.messageBar()):
-                self.build_layer(rlz_or_stat=self.gmpe_to_rlz[gmpe],
+                self.build_layer(rlz_or_stat=self.gmpe2rlz[gmpe],
                                  gmpe=gmpe)
                 self.style_maps()
         if self.npz_file is not None:
             self.npz_file.close()
 
-    def build_layer_name(self, gmpe, **kwargs):
+    def build_layer_name(self, gmpe=None, **kwargs):
         self.imt = self.imt_cbx.currentText()
         self.eid = self.eid_sbx.value()
         self.default_field_name = '%s-%s' % (self.imt, self.eid)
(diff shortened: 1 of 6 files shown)


svir/metadata.txt
```diff
@@ -23,6 +23,8 @@ email=staff.it@globalquakemodel.org
 
 # Uncomment the following line and add your changelog entries:
 changelog=
+    3.1.1
+    * When loading Ground Motion Fields, the names of the GMPEs are displayed instead of the realization names
```
micheles: When loading Ground Motion Fields -> When loading Ground Motion Fields for scenario calculations

svir/dialogs/load_gmf_data_as_layer_dialog.py
```diff
@@ -64,8 +64,39 @@ def __init__(self, iface, viewer_dock, session, hostname, calc_id,
...
+        self.show_num_sites()
+
+    def _format_rlz(self, gmpe_idx):
+        # NOTE: assuming that rlzs are always in the format rlz-XXX
+        return "rlz-%03d" % gmpe_idx
+
+    def show_num_sites(self):
+        # NOTE: we are assuming all realizations have the same number of sites,
+        #       which currently is always true.
+        #       If different realizations have a different number of sites, we
```
micheles: which currently is always true -> which is always true for scenario calculations

svir/dialogs/load_gmf_data_as_layer_dialog.py
```diff
@@ -64,8 +64,39 @@ def __init__(self, iface, viewer_dock, session, hostname, calc_id,
...
         self.ok_button.setEnabled(self.imt_cbx.currentIndex() != -1)
 
-    def on_rlz_or_stat_changed(self):
-        self.dataset = self.npz_file[self.rlz_or_stat_cbx.currentText()]
+    def populate_out_dep_widgets(self):
+        self.populate_gmpe_cbx()
+        self.show_num_sites()
+
+    def _format_rlz(self, gmpe_idx):
+        # NOTE: assuming that rlzs are always in the format rlz-XXX
```
micheles: Please use the name `rlz_idx`, not `gmpe_idx`

svir/dialogs/load_gmf_data_as_layer_dialog.py
```diff
@@ -64,8 +64,39 @@ def __init__(self, iface, viewer_dock, session, hostname, calc_id,
...
+    def populate_gmpe_cbx(self):
+        self.gmpes_npz = extract_npz(
+            self.session, self.hostname, self.calc_id, 'realizations',
+            message_bar=self.iface.messageBar(), params=None)
+        self.gmpes = self.gmpes_npz['array']['gsims']
+        self.gmpe_cbx.clear()
+        self.gmpe_cbx.setEnabled(True)
+        self.gmpe_cbx.addItems(self.gmpes)
+        self.rlzs_or_stats = [
+            self._format_rlz(gmpe_idx) for gmpe_idx in range(len(self.gmpes))]
```
micheles: Do we need `_format_rlz` ? Just write ['rlz-%03d'  % r for r in range(len(rlzs_npz))`

svir/dialogs/load_gmf_data_as_layer_dialog.py
```diff
@@ -64,8 +64,39 @@ def __init__(self, iface, viewer_dock, session, hostname, calc_id,
...
+        #       which currently is always true.
+        #       If different realizations have a different number of sites, we
+        #       need to move this block of code inside on_rlz_or_stat_changed()
+        rlz = self._format_rlz(self.gmpe_cbx.currentIndex())
+        gmpe_data = self.npz_file[rlz]
+        self.num_sites_lbl.setText(
+            self.num_sites_msg % gmpe_data.shape)
+
+    def populate_gmpe_cbx(self):
+        self.gmpes_npz = extract_npz(
```
micheles: The name must be rlzs_npz, NOT gmpe_rlzs. There is not one-to-one correspondence to realizations and GMPEs, except accidentally in the scenario calculator.

svir/dialogs/load_gmf_data_as_layer_dialog.py
```diff
@@ -64,8 +64,39 @@ def __init__(self, iface, viewer_dock, session, hostname, calc_id,
...
+        self.rlzs_or_stats = [key for key in sorted(self.npz_file)
+                              if key not in ('imtls', 'array')]
+        self.rlzs_npz = extract_npz(
+            self.session, self.hostname, self.calc_id, 'realizations',
+            message_bar=self.iface.messageBar(), params=None)
+        # NOTE: gmpe and gsim are synonyms
+        self.gmpes = self.rlzs_npz['array']['gsims']
+        assert len(self.rlzs_or_stats) == len(self.gmpes), (
+            "For scenario calculations, the number of realizations should be"
+            " equal to the number of GMPEs.\nRealizations: %s\nGMPEs: %s"
```
micheles: This is not the assert you want. You want `assert len(self.rlzs_or_stats) == len(set(self.gmpes))`  ;-)

svir/dialogs/load_gmf_data_as_layer_dialog.py
```diff
@@ -64,8 +64,40 @@ def __init__(self, iface, viewer_dock, session, hostname, calc_id,
...
         self.ok_button.setEnabled(self.imt_cbx.currentIndex() != -1)
 
+    def show_num_sites(self):
+        # NOTE: we are assuming all realizations have the same number of sites,
+        #       which is always true for scenario calculations.
+        #       If different realizations have a different number of sites, we
+        #       need to move this block of code inside on_rlz_or_stat_changed()
+        rlz = self.rlz_or_stat_cbx.itemData(
+            self.rlz_or_stat_cbx.currentIndex())
+        gmpe_data = self.npz_file[rlz]
```
micheles: gmpe_data is the wrong name. If I am not wrong, npz_file contains the GMFs, so the name should be gmf_data

svir/dialogs/load_gmf_data_as_layer_dialog.py
```diff
@@ -64,8 +64,40 @@ def __init__(self, iface, viewer_dock, session, hostname, calc_id,
...
+            self.num_sites_msg % gmpe_data.shape)
+
+    def populate_rlz_or_stat_cbx(self):
+        self.rlzs_or_stats = [key for key in sorted(self.npz_file)
+                              if key not in ('imtls', 'array')]
+        self.rlzs_npz = extract_npz(
+            self.session, self.hostname, self.calc_id, 'realizations',
+            message_bar=self.iface.messageBar(), params=None)
+        # NOTE: gmpe and gsim are synonyms
+        self.gmpes = self.rlzs_npz['array']['gsims']
```
micheles: You could use gsims everywhere, for consistency.

svir/dialogs/load_gmf_data_as_layer_dialog.py
```diff
@@ -64,8 +64,40 @@ def __init__(self, iface, viewer_dock, session, hostname, calc_id,
...
+        assert len(self.rlzs_or_stats) == len(set(self.gmpes)), (
+            "For scenario calculations, the number of realizations should be"
+            " equal to the number of distinct GMPEs.\n"
+            "Realizations: %s\nGMPEs: %s"
+            % (self.rlzs_or_stats, self.gmpes))
```
micheles: If the user see this, there is nothing she can do. This is actually a bug in the engine. Perhaps we can remove this assert here and add a check in the engine, so that a computation does not even start if we are in this condition.

PR #442, opened 2018-04-19
Add dmg_by_event and losses_by_event to outputs loaded from basic csv

ptormene:
Companion of https://github.com/gem/oq-engine/pull/3632

This is good as long as tables are small enough. Otherwise, we need to use a more efficient approach, leveraging the OQ-Engine 'extract' api, perhaps also checking the size of the output before attempting to load it into a QGIS layer.

NOTE: It does not work in case there are multiple realizations ==> multiple csv files zipped together


CatalinaYepes:
Depending on the calculation the output may have a .zip with multiple csv files. Currently the plugin only support csv files. We need to provide the option of selecting the realization, or remove the option to upload these type of layers.

ptormene:
Therefore, I prefer to close this PR. We will add loaders for these outputs that leverage the `extract` API instead of loading from CSV.

PR #447, opened 2018-04-27
Make versions more visible and stored in layers

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/445

screenshot from 2018-04-27 09-46-26

@pslh: I'm displaying the hostname instead of the name of the connection profile. This is easier to implement, and it looks good enough for me.


pslh:
The logo positions look good and I agree that the connection URL is sufficient for now - the connection name would be better, but I do not see this as a show-stopper.
LGTM.

PR #455, opened 2018-05-29
Remove SAGA

ptormene:
SAGA is not reliable across platforms, so we decided to use only the fallback algorithm and to completely remove the dependency from SAGA. In most cases, the internal algorithm is fast enough. About 200 lines of code were removed, and complexity was reduced. In a separate PR, I will leverage the processing toolbox to further simplify the code, letting QGIS perform the task (more efficiently).

I've merged also the changes made by @daniviga, using Bionic instead of Xenial for Travis (without installing SAGA in the docker)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -21,8 +21,7 @@ RUN apt update && apt -q install -y gnupg2 && \
 # a new 'apt update' is always performed to comply
 # with the cache invalidation
 ADD http://qgis.org/debian-ltr/dists/bionic/main/binary-amd64/Packages /tmp/nocache.json
-RUN apt install -y qgis python-qgis \
-                   saga python-saga && \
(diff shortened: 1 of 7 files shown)


docker/Dockerfile
```diff
@@ -1,22 +1,27 @@
...
 
-RUN add-apt-repository -y ppa:openquake/saga && \
-    apt update && \
-    apt install -y qgis python-qgis \
-                   saga python-saga && \
+# Perform cache invalidation when new packages are out
+# a new 'apt update' is always performed to comply
+# with the cache invalidation
+ADD http://qgis.org/debian-ltr/dists/bionic/main/binary-amd64/Packages /tmp/nocache.json
+RUN apt install -y qgis python-qgis && \
```
daniviga: We must keep the `apt update` here (my fault)

PR #456, opened 2018-06-05
Use a "processing" algorithm to aggregate points by zone

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/452

The Processing toolbox was improved, and it's said to perform now better than SAGA,  and the dependency from SAGA has always been problematic to maintain. Therefore I decided to switch to an algorithm called "Join attributes by location (summary)" instead of leveraging SAGA or running our own fallback implementation. I fixed the workflow aggregating losses or damages while loading outputs from the OQ-Engine, accordingly.
Instead of fixing the recovery modeling workflow the same way, I decided to remove recovery modeling completely from the plugin. It might be re-added in the future, perhaps as a separate plugin. Removing it, I've also removed the dependency from Scipy, that was used only there.
We also had a specific dialog to load a couple of layers and perform the aggregation of points by zone. Since the same can be done through Processing, it became useless to keep maintaining a dialog that does the same (and with less features). Therefore, I removed also that.
The portions of code that were deleted were quite complicated, so the result of this refactoring is a significant reduction of complexity in the code base.
In tests for aggregations I converted some reference shapefiles into geopackages. It avoids issues related to shapefile limitations. Unfortunately, it looks like QGIS modifies geopackages when opening them, regardless from the fact they are actually edited/saved or simply opened and read. As an ugly workaround, the test makes copies of those files and uses copies instead of the original files. This prevents git from notifying changes in the files after running tests.

TODO (in separate PRs):
- [ ] There's no feedback while aggregating points by zone. It has to be managed properly (perhaps @mbernasocchi can help me on this)


ptormene:
Replaced by https://github.com/gem/oq-irmt-qgis/pull/469

PR #467, opened 2018-06-27
Fix importing matplotlib, using Qt4Agg

ptormene:
It should solve the `tkinter` issue on Windows 64bit
Fixes https://github.com/gem/oq-irmt-qgis/issues/466


daniviga:
Please update the documentation

PR #468, opened 2018-06-27
Fix matplotlib, using Qt5Agg

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/466

### DIFF
--- a/svir/recovery_modeling/recovery_modeling.py
+++ b/svir/recovery_modeling/recovery_modeling.py
@@ -26,6 +26,8 @@ import os
 import json
 import bisect
 try:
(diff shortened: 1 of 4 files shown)


svir/recovery_modeling/recovery_modeling.py
```diff
@@ -26,6 +26,8 @@
 import json
 import bisect
 try:
+    import matplotlib
+    matplotlib.use('Qt5Agg')
     import matplotlib.pyplot as plt
 except ImportError as exc:
```
daniviga: I would remove the exception

PR #478, opened 2018-07-16
Visualize loss curves aggregated by tag

ptormene:
Companion of https://github.com/gem/oq-engine/pull/3862
Fixes https://github.com/gem/oq-irmt-qgis/issues/463

### DIFF
--- a/svir/dialogs/viewer_dock.py
+++ b/svir/dialogs/viewer_dock.py
@@ -373,6 +373,24 @@ class ViewerDock(QDockWidget, FORM_CLASS):
             return
         self.draw_dmg_by_asset_aggr()
 
+    def filter_agg_curves(self):
+        params = {}
+        for tag_name in self.tags:
+            if self.tags[tag_name]['selected']:
+                for value in self.tags[tag_name]['values']:
+                    if self.tags[tag_name]['values'][value]:
+                        # NOTE: this would not work for multiple values per tag
+                        params[tag_name] = value
+        to_extract = 'agg_curves/%s' % self.loss_type_cbx.currentText()
+        with WaitCursorManager(
+                'Extracting...', message_bar=self.iface.messageBar()):
+            self.agg_curves = extract_npz(
+                self.session, self.hostname, self.calc_id, to_extract,
+                message_bar=self.iface.messageBar(), params=params)
+        if self.agg_curves is None:
+            return
+        self.draw_agg_curves(self.output_type)
+
     def filter_losses_by_asset_aggr(self):
         params = {}
         for tag_name in self.tags:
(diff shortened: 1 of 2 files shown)


svir/dialogs/viewer_dock.py
```diff
@@ -373,6 +373,24 @@ def filter_dmg_by_asset_aggr(self):
             return
         self.draw_dmg_by_asset_aggr()
 
+    def filter_agg_curves(self):
+        params = {}
+        for tag_name in self.tags:
+            if self.tags[tag_name]['selected']:
+                for value in self.tags[tag_name]['values']:
+                    if self.tags[tag_name]['values'][value]:
+                        # NOTE: this would not work for multiple values per tag
```
micheles: It would be useful in a comment to write down the structure of the .tags dictionary in the example of the demo.

svir/dialogs/viewer_dock.py
```diff
@@ -373,6 +373,24 @@ def filter_dmg_by_asset_aggr(self):
...
+                        params[tag_name] = value
+        to_extract = 'agg_curves/%s' % self.loss_type_cbx.currentText()
+        with WaitCursorManager(
+                'Extracting...', message_bar=self.iface.messageBar()):
+            self.agg_curves = extract_npz(
+                self.session, self.hostname, self.calc_id, to_extract,
+                message_bar=self.iface.messageBar(), params=params)
+        if self.agg_curves is None:
```
micheles: In which case this happens?

svir/dialogs/viewer_dock.py
```diff
@@ -1430,24 +1481,32 @@ def write_export_file(self, filename):
...
                 writer.writerow(headers)
-                for i, return_period in enumerate(
+                for return_period_idx, return_period in enumerate(
                         self.agg_curves['return_periods']):
-                    values = self.agg_curves['array'][self.current_loss_type]
                     row = [return_period]
-                    row.extend([value for value in values[i]])
+                    for stat_idx, stat in enumerate(stats):
+                        row.append(self.agg_curves[
+                            'array'][stat_idx][return_period_idx])
```
micheles: This should work:
```
[stat_idx][return_period_idx] -> [stat_idx, return_period_idx] ?
```

svir/dialogs/viewer_dock.py
ptormene: `self.tags` contains a lot of tags and corresponding values. It's too big to be written in a comment here. I can edit it and reduce it to a very small subset, in order to make it readable here.

svir/dialogs/viewer_dock.py
ptormene: I think you are right. This should never happen, or it should break in case it happens.

ptormene: I've added a check while drawing, that clears the plot in case too much filtering produces an empty array.

svir/dialogs/viewer_dock.py
ptormene: Thanks

PR #488, opened 2018-08-01
Add checkbox to show return time; modify layer names

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/487


daniviga:
Can't this be applied also for UHS?

we have `uhs_poe-0.1_50.0y` that can be `500yr`, no?

PR #491, opened 2018-08-07
Load hazard curves for selected realization/statistic and intensity measure type

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/485


micheles:
Notice that since the feature of downloading only a specific IMT/statistic is missing server side, this feature is useless performance-wise. Behind the scenes, the pluging downloads everything.

ptormene:
Yes, ok, but we want to be able to load only the selected data, instead of loading everything and having to delete columns afterwards. It's not just a matter of performance.

PR #502, opened 2018-09-10
Upload project from qgis2 to new platform

hascar:
Upload project from qgis2 to new platform with correct style


svir/dialogs/download_layer_dialog.py
```diff
@@ -120,11 +120,17 @@ def get_capabilities(self):
     def parse_get_capabilities(self, xml):
         # this raises a IOError if the file doesn't exist
         root = ElementTree.fromstring(xml)
-        layers = root.find('%sFeatureTypeList' % NS_NET_OPENGIS_WFS)
+        # import pdb; pdb.set_trace()
+        featuretypelist = root.find('%sFeatureTypeList' % NS_NET_OPENGIS_WFS)
+        # typelist = root.find('{http://www.opengis.net/wfs}FeatureTypeList')
+        layers = featuretypelist.findall('%sFeatureType' % NS_NET_OPENGIS_WFS)
+        # layers = root.find('%sFeatureType' % NS_NET_OPENGIS_WFS)
```
ptormene: Please remove comments, unless they are actually useful

svir/dialogs/download_layer_dialog.py
```diff
@@ -120,11 +120,17 @@ def get_capabilities(self):
...
 
         for layer in layers:
             try:
                 keywords = layer.find('%sKeywords' % NS_NET_OPENGIS_WFS).text
+                title = layer.find('%sTitle' % NS_NET_OPENGIS_WFS).text
+                print("%s: %s" % (title, keywords))
```
ptormene: Please remove debug prints

svir/thread_worker/upload_worker.py
```diff
@@ -84,12 +84,36 @@ def work(self):
...
+                                     'change_layer_style', 'add_layer',
+                                     'change_layer', 'delete_layer',
+                                     'view_resourcebase',
+                                     'download_resourcebase',
+                                     'change_resourcebase_metadata',
+                                     'change_resourcebase',
+                                     'delete_resourcebase',
+                                     'change_resourcebase_permissions',
+                                     'publish_resourcebase',
+                                  ]
```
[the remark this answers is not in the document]
ptormene: In case you actually need the same list of permissions in both cases, I suggest to define the common list before defining `permissions`, and then to use it, therefore avoiding a dozen redundant lines of code.

svir/dialogs/upload_dialog.py
```diff
@@ -137,11 +137,15 @@ def _update_layer_style(self):
             os.close(fd)
             with open(fname, 'w') as f:
                 f.write(sld)
-            os.system('tidy -xml -i %s' % fname)
+            # os.system('tidy -xml -i %s' % fname)
         headers = {'content-type': 'application/vnd.ogc.sld+xml'}
+        resp = self.session.post(
```
ptormene: Please add a comment here, clarifying why we are doing both a post and a put. For instance: `NOTE: to save the style, we actually must do both the post and the put in this sequence, otherwise it does not work. We still haven't found any better way to obtain the same result.`

svir/dialogs/upload_dialog.py
```diff
@@ -139,9 +139,13 @@ def _update_layer_style(self):
                 f.write(sld)
             os.system('tidy -xml -i %s' % fname)
         headers = {'content-type': 'application/vnd.ogc.sld+xml'}
+        resp = self.session.post(
```
ptormene: Please add a comment like this: ```
NOTE: to save the style, we actually need to do both the post and put in this sequence, otherwise it doesn't work. We still haven't found any better way to obtain the same correct behavior
```

PR #515, opened 2018-09-28
Add the possibility to abort a OQ-Engine calculation

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/513


micheles:
I just tried it on my laptop. Works as expected.

PR #520, opened 2018-10-04
Add instructions working around a bug in matplotlib 3.0.0 for macOS

ptormene:
Actually, with @raoanirudh, we tested the workaround on matplotlib v2.2.2, but I believe also v2.2.3 would work fine as well. We should double-check it though.


ptormene:
I've just tried upgrading matplotlib to v3.0.0 on Ubuntu, and everything seems to work. I am curious what happens on Fedora.

PR #529, opened 2018-12-31
Fix call to the extract api and log as info both url and params

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/525

While fixing that issue, I noticed a couple of bugs, that I've fixed too:

1. the log verbosity was not getting the information from the right variable, therefore it was always set to WARNING, regardless from what the user selected

2. the task that calls the extract api was ignoring parameters, so the outputs were wrong. I can't understand why I did not spot it before.

### DIFF
--- a/svir/metadata.txt
+++ b/svir/metadata.txt
@@ -25,6 +25,8 @@ email=staff.it@globalquakemodel.org
 changelog=
     3.3.0
     * Fixed loader for aggregated Average Asset Losses Statistics OQ-Engine output
+    * All calls to the OQ-Engine "extract" API are logged
+    * Fixed a bug that caused the log verbosity to be always set to "warning", instead of reading it correctly from user settings 
     3.2.12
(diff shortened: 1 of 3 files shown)


micheles:
But where do we use parameters? I thought the calls to the WebUI did not need query parameters. Apart from that, LGTM and Happy New Year!

ptormene:
I've just checked about parameters. Actually, things were working correctly, because we were never using parameters in any of the calls to ExtractNpzTask, but only in calls to the extract_npz utility. For example, parameters are used while visualizing aggregate Average Asset Losses for a Scenario Risk calculation (e.g. GET: http://localhost:8800/v1/calc/25/extract/agg_losses/nonstructural, with parameters: {'NAME_1': 'Central'}). The extract_npz utility is used for calls that are supposed to be quick enough to be done synchronously, so it doesn't use a separate task.
Although we are not using parameters in ExtractNpzTask (yet), I prefer to make it possible, for the future. I am going to change a very little detail that I forgot, then it should be ready.

PR #540, opened 2019-02-26
Let integration tests reuse more code from the plugin instead of reimplementing parts of it

ptormene:
Some functionalities of the dialog that drives the Oq-Engine were re-implemented in the integration tests instead of leveraging the dialog itself.

I still need to fix the cases in which an output is downloaded from the Oq-Engine, for which we should test the actual downloader task. It was tricky and I couldn't fix that quickly, so I will postpone it to a separate PR.

### DIFF
--- a/svir/dialogs/drive_oq_engine_server_dialog.py
+++ b/svir/dialogs/drive_oq_engine_server_dialog.py
@@ -128,7 +128,7 @@ class DriveOqEngineServerDialog(QDialog, FORM_CLASS):
     it is possible to run calculations, delete them, list them, visualize
     their outputs and loading them as vector layers.
     """
-    def __init__(self, iface, viewer_dock):
+    def __init__(self, iface, viewer_dock, hostname=None):
         self.iface = iface
         self.viewer_dock = viewer_dock  # needed to change the output_type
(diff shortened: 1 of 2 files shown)


svir/dialogs/drive_oq_engine_server_dialog.py
```diff
@@ -140,7 +140,11 @@ def __init__(self, iface, viewer_dock):
         # keep track of the log lines acquired for each calculation
         self.calc_log_line = {}
         self.session = None
-        self.hostname = None
+        self.hostname = hostname
+        if hostname is not None:
+            self.forced_hostname = True
+        else:
```
micheles: Why not just use the hostname in the `ifs` ? The flag forced_hostname does not do much,

ptormene: It's a bit tricky. While playing with the plugin, the user might use the plugin settings dialog to set up the connection to a different engine server. If it happens, then when the dialog that drives the engine attempts to login to the engine server, we must get the modified user credentials (see below).
I am adding the hostname parameter here to be able to call this from tests, forcing the hostname to remain as specified, instead of reading it from the user settings.

PR #541, opened 2019-03-08
Show traceback

ptormene:
This is an improvement to the utility we use in the plugin to communicate with the user.
In case `log_msg` is called passing also an exception as parameter, the user will be able to visualize the traceback by clicking on a button that will appear in the messageBar. Therefore the messageBar will display the short version of the error message, but it will be easy to visualize the full error. 

To see how it works, you can attempt to drive the engine while the engine server is not running.


micheles:
LGTM

PR #542, opened 2019-03-15
In uhs integration test, accept more poe values

ptormene:


svir/test/integration/test_drive_oq_engine.py
```diff
@@ -179,10 +179,18 @@ def on_init_done(self, dlg):
         # set dialog options and accept
         if dlg.output_type == 'uhs':
             dlg.load_selected_only_ckb.setChecked(True)
-            idx = dlg.poe_cbx.findText('0.1')
-            assert_and_emit(dlg.loading_exception, self.assertEqual,
-                            idx, 0, 'POE 0.1 was not found')
-            dlg.poe_cbx.setCurrentIndex(idx)
+            poe_guessed_values = ['0.1', '0.02']
+            poe_was_found = False
```
micheles: This approach is rather fragile

ptormene: I will open a ticket, to make this and other checks less fragile. Instead of expecting a precise value, I can expect that the dropdown menu contains at least one element, and attempt using it.

ptormene: (or perhaps it's quicker to just do it now)

ptormene: Done. I think it's better now.

PR #543, opened 2019-03-26
Fixed the management of realizations

micheles:
Companion of https://github.com/gem/oq-engine/pull/4569


micheles:
LGTM if tests are green

PR #562, opened 2019-04-30
Fix newline separators in csv files exported by the Data Viewer

ptormene:
csv.writer was already producing data with the \r\n separator, but the problem was in the additional lines on top of the files, prepended by a #, that contain metadata such as the creation time, the engine and plugin version and some other scientific data.
Now I am using \r\n as line separator also while writing those headlines.

Fixes https://github.com/gem/oq-irmt-qgis/issues/561


micheles:
LGTM

PR #566, opened 2019-05-03
Newline

ptormene:
From the manual of python builtin `open` function:
 
>* On input, if newline is None, universal newlines mode is
>    enabled. Lines in the input can end in '\n', '\r', or '\r\n', and
>    these are translated into '\n' before being returned to the
>    caller. If it is '', universal newline mode is enabled, but line
>    endings are returned to the caller untranslated. If it has any of
>    the other legal values, input lines are only terminated by the given
>    string, and the line ending is returned to the caller untranslated.
    
>* On output, if newline is None, any '\n' characters written are
>    translated to the system default line separator, os.linesep. If
>    newline is '' or '\n', no translation takes place. If newline is any
>    of the other legal values, any '\n' characters written are translated
>    to the given string.

If the default is to translate all newlines to `os.linesep`, and if I understand correctly the consequences of it, it looks like we should explicitly use `newline=''`. in order to prevent from the automatic translation.

### DIFF
--- a/svir/dialogs/download_layer_dialog.py
+++ b/svir/dialogs/download_layer_dialog.py
@@ -269,7 +269,7 @@ class DownloadLayerDialog(QDialog, FORM_CLASS):
                 raise SvNetworkError(get_style_resp.reason)
             fd, sld_file = tempfile.mkstemp(suffix=".sld")
             os.close(fd)
-            with open(sld_file, 'w') as f:
+            with open(sld_file, 'w', newline='') as f:
                 f.write(get_style_resp.text)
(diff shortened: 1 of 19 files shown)


ptormene:
We tried to write a csv file on Windows with and without `newline=''`, and then to read it on Ubuntu.
Without `newline=''`, on windows a second `\r' is added. Passing the same file on Ubuntu and reading it without `newline=''`, it causes rows to be separated by an extra newline.
Therefore, it looks like the safest thing to do is to always specify `newline=''` whenever writing or reading to file.

PR #577, opened 2019-05-16
Set dialogs to load outputs as layers, using a scrollable area for selectors

ptormene:
It should make dialogs usable also on low resolution screens


daniviga:
LGTM

Yes I agree with Cata, or otherwise it would be good also if they are a collapsible area

PR #583, opened 2019-06-11
Make the IRMT also a Processing Toolbox algorithm provider

ptormene:
I have added to the Processing Toolbox two groups of algorithms.

One, "Zonal aggregation", contains only one element: "Join attributes by location (summary) and style output", which is a copy of an already available algorithm, plus the possibility to style the output by one of its fields. Unfortunately, I was unable to subclass the original algorithm and just add a postProcessing method to style the output; I would appreciate support from @mbernasocchi and @marioba about this point.

The other group, "Field transformation", contains all the transformation algorithms that were already available in the plugin outside the Processing Toolbox, and that were accessible via the plugin menu and toolbar, through a dialog. I still have not removed the dedicated dialog, because it provides some additional features that I have not made available in Processing: 1) the possibility to overwrite the original fields with the transformed ones; 2) the possibility to update project definitions pointing to the transformed fields instead of the original ones.
Perhaps it could be a good idea to extract the data transformation algorithms into a separate module, that would become another plugin from which the IRMT would depend. It is quite general purpose and also easy to extend adding other algorithms to transform vector layer fields. To add a new algorithm it would be sufficient to subclass `TransformFieldsAlgorithm`, re-implementing just a few methods and calling a corresponding transformation function.

I have slightly modified the module that contains the transformation functions, in order to make it work both with the old dialog and with the Processing Toolbox. The idea is to keep testing the original functions, while using them in different tools.

I have also added to the user manual a short note mentioning the new algorithm provider. Furthermore, in the `shortHelpString` of each transformation I have copied/adapted/improved from the user manual the description of the corresponding algorithm and its parameters, which makes the tool more user-friendly.


svir/calculations/process_layer.py
```diff
@@ -379,8 +379,10 @@ def transform_attribute(
             for feat in self.layer.getFeatures(request):
                 feat_id = feat.id()
                 value = transformed_dict[feat_id]
-                if type(value) not in (NULL, type(None)):
+                try:
                     value = float(value)
+                except Exception:
```
marioba: I think is better to handle only `ValueError` exception instead of handling generic exceptions

PR #590, opened 2019-06-21
WIP: Add manual section about CRAVE, with a set of basic examples

ptormene:
I've added some simple examples that should cover most of the cases that we might need (lists, cross-references, images, blocks, bibliography, etc.

This could be our starting point.


ptormene:
Replaced by https://github.com/gem/crave-docs

PR #594, opened 2019-07-04
Major improvements in integration tests

ptormene:
Thanks to the QGIS docker, we can now remove every mocked objects from integration tests.
I've also noticed that in many tests I was selecting all the features and drawing all the corresponding curves, whereas it is sufficient to draw just a couple. This change made the tests for many output loaders much faster. In fact, it reduced the execution of integration tests from 658.462s to 236.349s, and we don't have any warnings about loaders taking more than 10 seconds to run.
I changed the way loaders are tested. Instead of running a single test with a big loop, trying to load all the outputs of all the available calculations, I have now one test for each loader, looping on all the available calculations and loading a specific output type for all of them. Each test produces its own summary, and there is a general summary once all tests are finished.
I also added new tests that became necessary after this change. One checks that all loadable output types appear at least once in the OQ-Engine demos (besides those with the _aggr suffix, that are treated differently). The other one lists all the output types for which we have implemented a loader, and those for which a loader is still missing (the test will always be green anyway, but it provides useful information).
I've also added a test running demo calculations (a hazard one, then a risk one on top of the former) and finally deleting the added calculations: https://github.com/gem/oq-irmt-qgis/pull/596

### DIFF
--- a/scripts/run_integration_tests.sh
+++ b/scripts/run_integration_tests.sh
@@ -2,14 +2,29 @@
 
 set -e
 
+if [ "$TRAVIS_PULL_REQUEST_BRANCH" != "" ]; then
+    BRANCH=$TRAVIS_PULL_REQUEST_BRANCH
+elif [ "$TRAVIS_BRANCH" != "" ]; then
+    BRANCH=$TRAVIS_BRANCH
+else
+    BRANCH='master'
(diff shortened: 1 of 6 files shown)


micheles:
LGTM

PR #595, opened 2019-07-09
Disable whatever calls extract/agg_curves

ptormene:
Companion of https://github.com/gem/oq-engine/pull/4882

Besides some code re-alignment, the only actual change is commenting out the loader for `agg_curves-stats`, that is the only one that uses `extract/agg_curves/loss_type`


micheles:
LGTM

PR #598, opened 2019-07-11
Hovering on the name of an engine calc output, a tooltip displays the output type and its size

ptormene:
(e.g. Aggregate Asset Losses -> agg_losses-rlzs)

tooltip

In the past, we decided to avoid displaying such information in the table, keeping the visualized data simpler and cleaner. However, this is visualized in the WebUI, and it can be useful in some cases.
I decided to implement this hybrid approach, to allow users to easily obtain the information by pointing the mouse to the output name and visualizing the tooltip, so the table remains as simple and clean as before.


ptormene:
In the WebUI, a tooltip shows the size of the output. I might add also that information to the tooltip I am adding here (e.g. "output_short_name (0.5 MB)"

ptormene:
tooltip2

PR #608, opened 2019-07-23
Load dmg_by_asset with or without aggregating by site

ptormene:
and use the output without aggregation by site to perform recovery modeling.

In order to make recovery modeling work, I transform the damage state values, that currently correspond to the number of buildings in each damage state (instead of being the probability for a building to be in each damage state). By dividing the number of building in each damage state by the total number of buildings, I obtain the probabilities summing to 1, that are usable by the recovery modeling tool.

I have also added integration tests to run the recovery modeling for 1 or 2 points in the map, in case of no aggregation by site, with different recovery time approaches.

I have added 2 checkboxes: one to force the selection of all the assets that are at the same coordinates of the one that is randomly selected by QGIS when you click on overlapping points; the second to enable/disable recalculating the recovery curve on-the-fly.

I removed functional tests completely. They included only tests about recovery modeling, and they assumed some reference data that became outdated and that would become outdated again each and every time the dmg_by_asset output changes its format. Instead of those precise but fragile tests, I have now some checks in the integration tests, to make sure that recovery curves are in the range 0-1, as well as normalized probabilities to be in each damage state.

TODO:

- [x] when 'load only the selected items' is unchecked, both the aggregated by site and the not aggregated kinds of layer have to be loaded correctly, with meaningful names


ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/609

ptormene:
Also fixes https://github.com/gem/oq-irmt-qgis/issues/249

PR #615, opened 2019-08-02
WIP: load agg_curves-stats reading from new oq-engine export format

ptormene:
Some specifications from @raoanirudh:
- [x] by default, for each tag name, preselect 1 tag value
- [ ] add the possibility to select multiple tags and multiple values per tag (PT: in order to obtain a 2D array that is suitable for plotting, we must force the selection of one and only one value for each tag. However, it is possible to make different selections, and then add the filtered data to the same plot)
- [ ] also for agg_curves-rlz add selectors for tags and their values with the same behavior
- [ ] when displaying plots for multiple tag values, they should have different colors and corresponding legend (we could use different symbols for realizations/stats, and different colors for tags (PT: I need to better investigate this aspect)


ptormene:
This PR was replaced by https://github.com/gem/oq-irmt-qgis/pull/628

PR #623, opened 2019-08-07
WIP: improve management of signals in integration tests

ptormene:
I am trying to better identify the objects that emit signals. I suspect this is necessary to avoid some issues that might occur if a slot receives a signal and performs an action as a consequence, but the signal was supposed to be used from a different context.


ptormene:
This was just an experiment that did not bring to the expected results, so I am closing the PR

PR #628, opened 2019-08-19
WIP: load agg_curves visualizing curves corresponding to multiple tag values

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/616

This already works for agg_curves-stats and agg_curves-rlzs, but the changes broke other loaders, that I am still fixing.

The functionality to export the visualized curves as csv also still needs to be fixed, but I would prefer to work on it after we agree if this approach is better with respect to the one implemented in the `ebrisk` branch


ptormene:
I have fixed what was broken for the other outputs, and the exporting functionalities. In case of exporting data for multiple tag values, the export could be improved.
I am still fixing synchronization issues in tests, especially with respect to tests for recovery modeling with various settings for parameters.

ptormene:
TODO:
[ ] for ebrisk agg_curves-rlzs I should have one monoselect combobox per tag

ptormene:
Also fixes https://github.com/gem/oq-irmt-qgis/issues/423

PR #629, opened 2019-08-19
Use ScenarioDamage risk test in integration tests

ptormene:
I was using a fragile approach to run the first found risk demo, but recently something changed and the test was not running as expected anymore. Another problem was that, in case of multiple ini files, the plugin now asks the user which one to use for the calculation, which would put the tests on hold, waiting for a response from the user. Now, while running tests, I use instead the original approach to let the engine guess which ini file to use from those that are available.


micheles:
Asking the user is good in interactive mode, while in test mode is good to use a default. I guess we can determine if we are in testing model by looking at the TRAVIS environment variable.

ptormene:
I am using an optional parameter (`use_default_ini`) that is `False` by default, but that is explicitly set to `True` when called from tests.

PR #647, opened 2019-09-10
Implement spatial join summary style algorithm as subclass

marioba:
Implement the processing algorithm "Join attributes by location (summary) and style output" as a subclass of the [QGIS `SpatialJoinSummary` processing algorithm](https://github.com/qgis/QGIS/blob/8a3c7b14c367771d096b4a6d006aa3c4b1017dd5/python/plugins/processing/algs/qgis/SpatialJoinSummary.py)


svir/processing_provider/spatial_join_summary_style.py
```diff
@@ -22,48 +22,20 @@
...
-    JOIN = "JOIN"
-    PREDICATE = "PREDICATE"
-    JOIN_FIELDS = "JOIN_FIELDS"
-    SUMMARIES = "SUMMARIES"
-    DISCARD_NONMATCHING = "DISCARD_NONMATCHING"
-    OUTPUT = "OUTPUT"
-    ADDED_FIELDS = "ADDED_FIELDS"
+class SpatialJoinSummaryStyle(SpatialJoinSummary):
+
+    def __init__(self):
```
ptormene: Perhaps this could be removed

svir/processing_provider/spatial_join_summary_style.py
```diff
@@ -114,321 +83,33 @@ def tr(self, string):
...
-                self.ADDED_FIELDS: self.fields_to_join}
+        input_layer = QgsProcessingUtils.mapLayerFromString(
+            self.parameters[self.INPUT], context)
 
-    def postProcessAlgorithm(self, context, feedback):
         processed_layer = QgsProcessingUtils.mapLayerFromString(
             self.dest_id, context)
-        added_fieldnames = [field.name() for field in self.fields_to_join]
+
+        added_fieldnames = set(processed_layer.fields().names()) - set(
```
ptormene: Please check if this does not preserve the order of fields. Perhaps it should be slightly better to do a little loop to preserve the order.

svir/processing_provider/spatial_join_summary_style.py
marioba: yes

svir/processing_provider/spatial_join_summary_style.py
marioba: done

PR #684, opened 2019-09-26
Fix loader for ruptures

ptormene:
First, I am making sure that the integration test fails because we are using tab as separator instead of comma. Then I am fixing the separator.

### DIFF
--- a/scripts/run_integration_tests.sh
+++ b/scripts/run_integration_tests.sh
@@ -21,6 +21,7 @@ docker run -d --name qgis -v /tmp/.X11-unix:/tmp/.X11-unix \
  -e OQ_ENGINE_HOST='http://172.17.0.1:8800' \
  -e BRANCH="$BRANCH" \
  -e SELECTED_CALC_ID="$SELECTED_CALC_ID" \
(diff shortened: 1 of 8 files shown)


ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/673
Fixes https://github.com/gem/oq-irmt-qgis/issues/682

scripts/run_integration_tests.sh
```diff
@@ -29,4 +30,4 @@ docker exec -it qgis sh -c "git clone -q -b $BRANCH --depth=1 https://github.com
 
 docker exec -it qgis sh -c "qgis_setup.sh svir"
 
-docker exec -it qgis sh -c "cd /tests_directory && SELECTED_CALC_ID=$SELECTED_CALC_ID qgis_testrunner.sh svir.test.integration.test_drive_oq_engine"
```
micheles: Just curious, why `SELECTED_CALC_ID=$SELECTED_CALC_ID` disappeared?

ptormene: Yes, that's why

PR #690, opened 2019-09-27
Keep all tag values selectors visible instead of using tabs

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/675


ptormene:
To fix:

- [ ] when a tag name is deselected, all the corresponding values should be automatically deselected too

PR #692, opened 2019-10-01
Unskip test running risk on top of hazard, then removing both calculations

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/688

In master I was keeping this skipped because it was causing some issues while running the risk part of the calculation and attempting to remove both calculations afterwards. We can see if we still have those connection issues and if meanwhile everything is solved engine-side about removing calculations.


micheles:
I am not sure about what `irmt.drive_oq_engine_server_dlg.remove_calc(risk_calc_id)` does, but you should try to run the tests with this branch after https://github.com/gem/oq-engine/pull/5149 enters.

PR #698, opened 2019-11-05
Avoid to print the traceback of connection errors in integration tests

ptormene:
I am also handling some additional error types


micheles:
The tests are green and nice, LGTM :-)

PR #700, opened 2019-11-06
Extract agg_curves instead of agg_curves-stats and agg_curves-rlzs

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/644

TODO:

- [x] Extract units (as soon as that key becomes available from the OQ-Engine)
- [x] Fix management of tags in ebrisk


ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/610

PR #701, opened 2019-11-08
Filter ruptures by minimum magnitude

ptormene:
Leveraging the change made in https://github.com/gem/oq-engine/pull/5279/files


ptormene:
For some reason, the size of the dialog used to select the field for styling and the minimum magnitude for filtering is not adjusted properly.

PR #710, opened 2019-11-29
Prefilter ground motion fields by event id

ptormene:
and enable their visualization for event-based calculations

TODO:
- [x] always call oqparam while loading any output from the engine
- [x] remove selector for rlzs, because given an eid it corresponds to one and only one rlz
- [x] remove checkbox to load all realizations and label for selector of realizations

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: trusty
+dist: xenial
 services: docker
 
 language: python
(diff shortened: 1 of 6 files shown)


ptormene:
While working on this, I noticed that we are still not visualizing dmg_by_event and losses_by_event for event based (I suppose for the same reason why we were not visualizing gmf_data for event based). Perhaps a pre-filtering would be useful also in those cases to avoid dealing with huge amounts of data?

PR #723, opened 2020-02-07
Decode data right after extraction

ptormene:
In most cases, it avoids to keep decoding objects afterwards. It does not improve things in cases such as: the extracted dictionary contains a key 'array' and the array has for instance a column with bytes. In that kind of situation, I am still decoding those bytes afterwards.

NOTE: this change highlighted a corner case in which the plugin-side complains that it is impossible to read pickled data with allow_pickle=False. I created a branch engine-side called `fix-extract-val-none` that should address also that corner case. The point seems to be that if in the extracted object there is a key for which the value is None, it would still export it with type object. Instead of None, I am using an empty list, and it seems to work as expected (although I am not sure if it is the right thing to do).


micheles:
Using an empty tuple instead of None is what I do.

ptormene:
This change does not improve the situation very much, and it actually introduces a number of potential issues, so we decided to keep things as they were.

PR #733, opened 2020-04-21
Fix loading GMFs

ptormene:
While making a recent change I did not notice I was modifying a dict while looping on its keys, which caused the loop to make an undesired additional iteration. As a consequence, an empty field with a strange name was added to the layer.

Fixes https://github.com/gem/oq-irmt-qgis/issues/730
Fixes https://github.com/gem/oq-irmt-qgis/issues/728

### DIFF
--- a/svir/dialogs/load_output_as_layer_dialog.py
+++ b/svir/dialogs/load_output_as_layer_dialog.py
@@ -23,6 +23,7 @@
 # along with OpenQuake.  If not, see <http://www.gnu.org/licenses/>.
 
 import os
+import copy
 from random import randrange
 from osgeo import ogr
 from qgis.core import (QgsVectorLayer,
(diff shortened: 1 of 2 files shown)


svir/dialogs/load_output_as_layer_dialog.py
```diff
@@ -480,9 +482,9 @@ def build_layer(self, rlz_or_stat=None, taxonomy=None, poe=None,
                 if field_name == self.default_field_name:
                     self.default_field_name = added_field_name
                 # replace field_name with the actual added_field_name
-                field_type = field_types[field_name]
-                del field_types[field_name]
-                field_types[added_field_name] = field_type
+                del modified_field_types[field_name]
+                modified_field_types[added_field_name] = field_type
```
micheles, line 486: Why the line `del modified_field_types[field_name]` ?

CatalinaYepes:
An additional issue is appearing when loading the gmfs for scenarios.

Independently from the selected event_id, always the rlz-0 name is displayed.

For example the picture below show the results for a calculation with a total of 60 events with 3 GMPEs (20 events per GMPE), but the name is the same in all cases.

Screen Shot 2020-04-24 at 5 06 00 PM

PR #742, opened 2020-09-22
Relevant events

ptormene:
Now the OQ-Engine return only the relevant events, so the list of event ids might likely have missing values, so we need to help the user identifying the closest event id with respect to the initially chosen one.
I am also fixing some minor aspects due to recent changes in the engine and in the numpy api.

### DIFF
--- a/svir/dialogs/load_gmf_data_as_layer_dialog.py
+++ b/svir/dialogs/load_gmf_data_as_layer_dialog.py
@@ -76,7 +76,8 @@ class LoadGmfDataAsLayerDialog(LoadOutputAsLayerDialog):
         self.events_npz = events_npz
         events = events_npz['array']
         if 'GEM_QGIS_TEST' in os.environ:
-            self.eid, ok = 0, True
+            self.eid = self.get_closest_element(self.eid, events['id'])
+            ok = True
         elif 'scenario' in self.calculation_mode:
             range_width = self.oqparam['number_of_ground_motion_fields']
             ranges = {}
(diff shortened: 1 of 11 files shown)


svir/dialogs/load_gmf_data_as_layer_dialog.py
```diff
@@ -69,11 +69,15 @@ def __init__(self, drive_engine_dlg, iface, viewer_dock, session, hostname,
             self.on_extract_error)
         QgsApplication.taskManager().addTask(self.extract_npz_task)
 
+    def get_closest_element(self, element, elements):
+        return min(elements, key=lambda x: abs(x - element))
+
```
micheles, line 74: The idea is good, but you need to use numpy for performance:
```
return elements[numpy.abs(elements - element).argmin()]        
```

svir/dialogs/load_gmf_data_as_layer_dialog.py
```diff
@@ -86,12 +90,42 @@ def get_eid(self, events_npz):
...
+                self.eid, ok = QInputDialog.getInt(
+                    self.drive_engine_dlg,
+                    "Select an event ID", "Range (%s - %s)" % (
+                        events[0]['id'], events[-1]['id']),
+                    self.eid, events[0]['id'], events[-1]['id'])
+                if not ok:
+                    self.reject()
+                    return
         if not ok:
             self.reject()
```
micheles, line 130: I do not like the duplication of logic

micheles:
The tests are red

svir/dialogs/load_gmf_data_as_layer_dialog.py
ptormene: Thanks. I updated it as suggested.

svir/dialogs/load_gmf_data_as_layer_dialog.py
ptormene: I've removed the duplication of logic

micheles:
LGTM!

PR #754, opened 2020-12-28
Load the old agglosses OQ-Engine outputs as agg_losses-rlzs and agg_losses-stats

ptormene:


svir/dialogs/load_csv_as_layer_dialog.py
```diff
@@ -69,11 +70,31 @@ def load_from_csv(self):
...
-        csv_path = self.path
+        if os.path.splitext(self.path)[1] == '.zip':
+            save_dest_dir = tempfile.mkdtemp()
+            # unzip and load all
+            with zipfile.ZipFile(self.path, 'r') as zip_ref:
+                zip_ref.extractall(save_dest_dir)
+            for i, csv_path in enumerate(os.listdir(save_dest_dir)):
+                full_csv_path = os.path.join(save_dest_dir, csv_path)
+                if os.path.isfile(full_csv_path):
+                    if i == 0:
```
micheles, line 81: Since this is only to be used for the statistics, I would search for 'mean' in csv_path and extract that file.
If missing, raise an error. Taking the first file is very dangerous because the order of the files returned by `os.listdir` is unspecified.

ptormene: My idea was to load all the csv files (for all statistics) and to visualize the attribute table of only one, while the others can be visualized by the user afterwards by selecting the corresponding layer. Instead of displaying the table for the first csv, I can look for 'mean' and automatically display that, while still loading also the others as layers.

micheles:
You have removed `svir/dialogs/load_basic_csv_as_layer_dialog.py`. Is it okay? I would have expected somebody to import that module.

ptormene:
`dialogs/load_csv_as_layer_dialog.py` and `dialogs/load_basic_csv_as_layer_dialog.py` had exactly the same content and only the former had any reference in the rest of the code, so I removed the duplicate. I was surprised when I spotted that. I guess in the past I was doing some renaming and I accidentally added a duplicate.

micheles:
I see, the LGTM in that case.

PR #766, opened 2021-02-19
Add check for 'skip CI'; correctly set branches for IRMT and Engine

ptormene:
Attempting to fix the behavior of CI both for PRs and for daily runs.

To fix: 
- [x] adding (skip CI) did not prevent it from running the job anyway.
- [ ] it looks like github.event.repository.default_branch does not retrieve any value, so I put `master` explicitly instead (it does not look so bad anyway, since we are likely going to keep using master as default branch both on IRMT and Engine

### DIFF
--- a/.github/workflows/qgis_383.yml
+++ b/.github/workflows/qgis_383.yml
@@ -11,6 +11,8 @@ jobs:
   
   Tests_and_docs:
     runs-on: ubuntu-18.04
+    if: "!contains(github.event.head_commit.message, 'skip ci')"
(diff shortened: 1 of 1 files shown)


vot4anto (changes requested):
GITHUB_HEAD_REF is not defined, instead of GITHUB_HD_REF:  ${{ github.head_ref }}

ptormene:
According to https://docs.github.com/en/actions/reference/environment-variables, GITHUB_HEAD_REF is one of the default environment variables. It looks like you can use them without needing to redefine them in the `env` section.

vot4anto (approved):
LGTM

PR #768, opened 2021-02-24
Restructure integration tests to have one test case per output type, plus other various tests

ptormene:
With these changes, tests can be run versus different QGIS docker versions (3_8_3, 3_16 and latest).
Some deprecated functions are replaced with newer ones only if QGIS is detected to be recent enough to have the new API available.
Recovery modeling (that was broken and which corresponding tests were skipped) was fixed and it is now tested as well.
Also fixes https://github.com/gem/oq-irmt-qgis/issues/753


micheles:
Paolo, you should abandon unittest and use pytest that is much better for what you are trying to do.

ptormene:
My problem here is that I am importing qgis.unittest, so I have to understand if I can use pytest instead in this case. I am going to investigate on this.

PR #786, opened 2021-09-27
Replaced output `agg_losses-stats` with aggrisk

micheles:
Companion of https://github.com/gem/oq-engine/pull/7227.


ptormene (approved):
LGTM if tests on QGIS LTR are green

PR #791, opened 2021-11-03
Test versus all qgis versions from 3.8 to 3.22

ptormene:
This is to investigate the "QGIS died on signal 11" error that occurs with the most recent QGIS docker images.


ptormene:
Thanks to this investigation, we identified the version of QGIS that introduced the issue.

PR #802, opened 2022-12-27
Improve makefile

ptormene:
Improvements:

- the documentation is built using a docker containing all the necessary dependencies
- the version is read from metadata instead of setting an environmental variable manually
- the documentation is set as LTS or latest depending on the 'experimental' flag read from metadata

To be checked by @vot4anto:

- the docker builds the documentation setting all files as owned by root, so it's not possible to edit the pdf of the manual afterwards without updating permissions with sudo. Is there a better way to do this?

NOTE: integration tests fail because this is aligned with Engine 3.11, but tests are versus Engine master

### DIFF
--- a/scripts/make_doc.sh
+++ b/scripts/make_doc.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-/usr/bin/docker rm -f qgis || true
+docker rm -f qgis || true
 docker run -d --name qgis -v /tmp/.X11-unix:/tmp/.X11-unix \
  -v `pwd`/../.:/oq-irmt-qgis \
  -e DISPLAY=:99 \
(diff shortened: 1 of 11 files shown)


scripts/make_doc.sh
```diff
@@ -2,12 +2,13 @@
 
 set -e
 
-docker rm -f qgis || true
+/usr/bin/docker rm -f qgis || true
```
ptormene, line 5: We can remove /usr/bin

scripts/make_doc.sh
```diff
@@ -2,12 +2,13 @@
...
 docker run -d --name qgis -v /tmp/.X11-unix:/tmp/.X11-unix \
  -v `pwd`/../.:/oq-irmt-qgis \
  -e DISPLAY=:99 \
  qgis/qgis:final-3_8_3
 
 docker exec -it qgis sh -c "apt update --allow-releaseinfo-change; DEBIAN_FRONTEND=noninteractive apt install -y latexmk texlive-latex-extra python3-matplotlib python3-sphinx python3-sphinx-rtd-theme dvipng"
 
-docker exec -it qgis sh -c "export PYTHONPATH=$PYTHONPATH:/oq-irmt-qgis; cd /oq-irmt-qgis/svir; make doc"
+# docker exec -it qgis sh -c "export PYTHONPATH=$PYTHONPATH:/oq-irmt-qgis; cd /oq-irmt-qgis/svir; make doc"
```
ptormene, line 13: We can remove the comment

svir/Makefile
```diff
@@ -117,8 +117,8 @@ package: compile doc
 	rm -f $(PLUGINNAME)/$(PLUGINNAME).zip
 	git archive --prefix=$(PLUGINNAME)/ -o $(PLUGINNAME).zip $(VERSION)
```
ptormene, line 118: Use HEAD instead of VERSION

vot4anto (approved):
LGTM

PR #803, opened 2022-12-27
Make Pillow not strictly required to install the plugin

ptormene:
UPDATE (after checking with AE): On wksptormene, if pillow is removed and matplotlib is kept, matplotlib complains about PIL missing. Therefore we should probably only check matplotlib and ignore pillow.

Pillow is used only by the social vulnerability widget, so it might be ignored by most of our current OQ users.
Therefore, I am now allowing the plugin to be installed even if this dependency is missing.

NOTE: integration tests fail because this is aligned with Engine 3.11, but tests are versus Engine master

### DIFF
--- a/svir/irmt.py
+++ b/svir/irmt.py
@@ -122,6 +122,7 @@ class Irmt(object):
             missing_packages.append('Pillow')
         if missing_packages:
             warn_missing_packages(missing_packages)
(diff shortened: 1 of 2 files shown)


ptormene:
Pillow turned out to be needed by some modules of matplotlib that we use in the dock window interface, so I prefer to postpone the changes proposed in this PR at least for now.

PR #828, opened 2023-06-07
Handle measurement units recently added to the engine: 'area' and 'number'

ptormene:
Companion of https://github.com/gem/oq-engine/pull/8838, contributes to https://github.com/gem/oq-engine/issues/8774

Should this change also be backported to the "stable" version of the plugin? I suppose we shouldn't, because 'area' and 'number' were introduced in https://github.com/gem/oq-engine/pull/8732, that is associated to the milestone `Engine 3.17.0`

For now, in case of displaying absolute values, we might have the case showing in the ylabel "Area loss (area)" that doesn't look optimal. However, if we find a way to provide engine-side the actual measurement unit, this code will remain valid and it will display something like "Area loss (sqm)" or "Area loss (sqft)".

The change in the unit test for loss aggregation is not actually needed. I did it just to have more consistency in the names we are using.


nicolepaul (approved):

svir/test/unit/test_loss_aggreagation.py
```diff
@@ -46,7 +46,7 @@ def setUp(self):
         curr_dir_name = os.path.dirname(__file__)
         self.data_dir_name = os.path.join(
             curr_dir_name, os.pardir, 'data', 'aggregation', 'dummy')
-        self.loss_attr_names = ['FATALITIES', 'STRUCTURAL']
+        self.loss_attr_names = ['PEOPLE', 'STRUCTURAL']
```
micheles, line 49: I see a confusion between loss type and unit of measure but not an easy solution (i.e. instead of PEOPLE we should have FATALITIES, INJURED or HOMELESS but then you would need to transmit the information from the engine to know what to use, while PEOPLE is good for everything).

ptormene: Anyway, this unit test is just checking that the plugin correctly aggregates "geographically" the values of a layer of points containing values of any kind, with respect to another layer containing polygons. I made that unneeded change only for the sake of naming consistency, but I could be aggregating anything.

PR #837, opened 2023-09-19
Adding OEP and AEP for event-based risk and damage calculations

ptormene:
Companion of https://github.com/gem/oq-engine/pull/9016
Addressing https://github.com/gem/oq-engine/issues/8971

Before the addition of the AEP and OEP curves, the plugin displayed multiple curves, one per each realization or statistic selected with a multi-select widget.
Now that also AEP and OEP curves are available, I've added a selector allowing to pick either `loss`, `loss_aep` or `loss_oep` and to display the corresponding curve for each of the realizations or statistics selected. The label for this selector is `Exceedance Probability`, but perhaps another better name could be given instead.

We may want to display AEP and OEP curves together instead of realizations/statistics. Or we may even want to display everything together (all realizations/statistics and all ep/aep/oep) in a plot that would become a bit noisy. In that case, labels and styles should be properly set for each curve.

My proposal is to merge into master the implemented approach, that fixes the integration tests and that is already enough to visualize all curves, even if perhaps in a sub-optimal way. We can further refine things in a later pull request.

### DIFF
--- a/svir/dialogs/viewer_dock.py
+++ b/svir/dialogs/viewer_dock.py
@@ -268,6 +268,18 @@ class ViewerDock(QDockWidget, FORM_CLASS):
         self.add_widget_to_type_dep_layout(
             self.imt_cbx, 'imt_cbx', self.typeDepHLayout1)
 
+    def create_ep_selector(self):
+        self.ep_lbl = QLabel('Exceedance Probability')
+        self.ep_lbl.setSizePolicy(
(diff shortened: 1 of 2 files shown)


micheles (approved):

ptormene:
I am merging this, so it will fix integration tests and it will make available the new functionalities, even if they require further improvements to be done in a separate PR.

PR #838, opened 2023-09-29
Increase the extent of layers having an empty extent in order to allow zooming to layer

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/206

I am doing it when creating layers of any kind, because the extent is always a rectangle and, even if in the properties it is displayed as empty, the object actually contains the coordinates. Therefore we can just increase the rectangle by a small delta.
With this fix, the plugin automatically (and correctly) zooms to layer also when loading a layer with only one point.

What we obtain is something more usable than before, although we introduce a little inconsistency, because we are declaring that the extent of our layer is bigger than its actual extent (that in fact is just a point). I believe the benefits justify the little inconsistency.


pslh (approved):
LGTM

ptormene:
I've ported this change also to v3.16.2 (stable) and to master and I've released v3.16.2 (stable) and v.3.17.8 (experimental).

PR #841, opened 2023-10-03
While extracting npz in numpy.load use max_header_size=100000 to fix e.g. some hcurves

ptormene:
Fixes https://github.com/gem/oq-irmt-qgis/issues/840

The stacktrace was misleading. Actually the plugin was unable to load hcurves from the npz successfully downloaded through the engine API, because numpy.load complained about max_header_size being too small for that specific calculation.
Raising that parameter default by 10 times, it should be sufficient in our use cases.
@alejocaldeGEM I've just tried also with the Croatia use case, successfully.

This change needs to be backported to 3.16 and 3.17.

### DIFF
--- a/svir/metadata.txt
+++ b/svir/metadata.txt
@@ -24,6 +24,7 @@ email=staff.it@globalquakemodel.org
 # Uncomment the following line and add your changelog entries:
 changelog=
     3.18.0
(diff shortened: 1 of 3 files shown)


ptormene:
The broken unit test on qgis latest docker is not related to this issue and it is caused by the sudden disappearing of the processing algorithm qgis:joinbylocationsummary, that is not documented and that might be an accidental deletion or a perhaps a replacement.

PR #842, opened 2023-10-03
Do not force min_mag=6.5 when loading ruptures in tests

ptormene:
This should avoid the following:
https://github.com/gem/oq-irmt-qgis/actions/runs/6387469847/job/17335848919#step:6:1546

See https://github.com/gem/oq-irmt-qgis/pull/843 about the broken unit test with QGIS LATEST.


svir/test/integration/test_drive_oq_engine.py
```diff
@@ -524,7 +524,7 @@ def load_output(
                 self.irmt.drive_oq_engine_server_dlg, self.irmt.iface,
                 self.irmt.viewer_dock,
                 self.irmt.drive_oq_engine_server_dlg.session,
-                self.hostname, calc_id, output_type, min_mag=6.5,
+                self.hostname, calc_id, output_type,
```
ptormene, line 527: Without specifying that parameter, the default is 4.0

PR #852, opened 2024-01-12
Extend matrix to run tests also on windows-latest and macos-latest

ptormene:


ptormene:
I am adding comments in a separate PR explaining why it does not work easily also on windows-latest and macos-latest

PR #872, opened 2024-10-03
Include the total losses in the selectable loss types when wisualizing aggcurves

ptormene:
Companion of https://github.com/gem/oq-engine/pull/10018

Fixes https://github.com/gem/oq-irmt-qgis/issues/783

### DIFF
--- a/svir/dialogs/viewer_dock.py
+++ b/svir/dialogs/viewer_dock.py
@@ -948,9 +948,9 @@ class ViewerDock(QDockWidget, FORM_CLASS):
 
         self.filter_avg_losses_rlzs_aggr()
 
-    def get_total_loss_unit(self, loss_type):
+    def get_total_loss_unit(self, total_losses):
         unit = None
-        separate_types = loss_type.split('+')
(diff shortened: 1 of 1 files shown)


svir/dialogs/viewer_dock.py
```diff
@@ -940,6 +948,28 @@ def load_avg_losses_rlzs_aggr(
...
+        for separate_type in separate_types:
+            if separate_type in self.single_loss_types:
+                loss_type_idx = self.loss_type_cbx.findText(separate_type)
+                this_unit = self.agg_curves['units'][loss_type_idx]
+                if unit is not None and unit != this_unit:
+                    raise ValueError(
+                        f'The unit of the total loss type ({loss_type}) can'
+                        f' not be determined by the separate loss types'
+                        f' ({separate_types}) because they have different'
+                        f' units ({unit} vs {this_unit})')
```
micheles, line 963: It makes little sense to do this check QGIS side, it is too late: if the units are inconsistent it must be an error engine side, even before starting the calculation.

PR #880, opened 2025-02-24
Update workflows

ptormene:
Use checkout@v4, setup-python@v5, python-version 3.12, remove inputs from workflow_dispatch and fix a typo


vot4anto (approved):
LGTM

PR #887, opened 2025-04-30
Load damages-stats similarly to damages-rlzs

ptormene:


svir/test/integration/test_drive_oq_engine.py
```diff
@@ -909,6 +919,14 @@ def test_load_damages_rlzs_aggr(self):
         self.load_output_type('damages-rlzs_aggr')
 
 
+class LoadDamagesStatsAggrTestCase(LoadOqEngineOutputsTestCase):
+    @unittest.skipIf(
+        ONLY_OUTPUT_TYPE and ONLY_OUTPUT_TYPE != 'damages-stats_aggr',
+        'only testing output type %s' % ONLY_OUTPUT_TYPE)
```
micheles, line 925: Same as before, it is not clear to me why we need the `skifIf`

ptormene: See above.

PR #907, opened 2026-04-28
Completely restructure integration tests to run without obsolete QGIS docker scripts

ptormene:
This reimplementation of tests was necessary, because so far we have been relying on scripts available in the QGIS docker images that have not been updated for years and that are no longer working with the most recent QGIS docker images.
With the new implementation, pytest fixtures are used to instantiate all the QGIS elements needed by the plugin, and tests are properly parametrized. This allowed to remove more than 400 hundred lines of code.

I will probably need to backport some changes to old branches.

I managed to make it work also for the workflow testing the compatibility between different releases of the plugin and of the engine:
https://github.com/gem/oq-irmt-qgis/actions/runs/25096016435

After merging this PR, I will update the qgis4 branch, trying to make tests work using the qgis4 docker image.


vot4anto (approved):
LGTM

PR #908, opened 2026-04-29
Upgrade the plugin and its tests to be compatible with QGIS4

ptormene:
We already did most of the functional migration, but the complex part was to upgrade the testing environment.
Now that https://github.com/gem/oq-irmt-qgis/pull/907 has been merged, I refreshed this branch.


ptormene:
For now I am keeping the qgis4 branch separated from master.

